PR #1, opened 2015-09-14
Add initial functions

paulcsmith:
* build/2
* create/2
* create_pair/2
* create_list/2
* assoc/2

### DIFF
--- a/lib/anvil.ex
+++ b/lib/anvil.ex
@@ -1,2 +1,63 @@
 defmodule Anvil do
+  defmodule UndefinedFactory do
+    defexception [:message]
+
+    def exception(factory_name) do
+      message = "No factory defined for #{inspect factory_name}"
+      %UndefinedFactory{message: message}
+    end
+  end
+
+  defmacro __using__(_opts) do
+    quote do
+      @before_compile unquote(__MODULE__)
+
+      def assoc(opts, factory_name) do
+        if Map.has_key?(opts, factory_name) do
+          Map.get(opts, factory_name)
+        else
+          create(factory_name)
+        end
+      end
+
+      def build(factory_name, opts \\ %{}) do
+        opts = Enum.into(opts, %{})
+        factory(factory_name, opts) |> Map.merge(opts)
+      end
+
+      def create(factory_name, opts \\ %{}) do
+        build(factory_name, opts) |> create_record
+      end
+
+      def create_pair(factory_name, opts \\ %{}) do
+        create_list(2, factory_name, opts)
+      end
+
+      def create_list(number_of_factorys, factory_name, opts \\ %{}) do
+        Enum.map(1..number_of_factorys, fn(_) ->
+          create(factory_name, opts)
+        end)
+      end
+    end
+  end
+
+  defmacro __before_compile__(_env) do
+    quote do
+      @doc """
+      Calls factory/1 with the passed in factory name
+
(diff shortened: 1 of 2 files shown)


jsteiner:
What do you think about making separate test files for `build`, `create`, etc?

test/anvil_test.exs
```diff
@@ -1,7 +1,131 @@
...
+  end
+
+  test "create/2 builds the fixture and saves it in the repo" do
+    MyApp.Anvil.create(:user, admin: true)
+
+    assert_received {:created, %{name: "John Doe", admin: true}}
+  end
+
+  test "create_pair/2 creates the fixture and saves it 2 times" do
+    MyApp.Anvil.create_pair(:user, admin: true)
```
jsteiner: should we assert that this actually returns a pair?

jsteiner: same with `create` and `create_list`

paulcsmith: Oh yes of course! Good catch

test/anvil_test.exs
```diff
@@ -1,7 +1,131 @@
...
-  test "the truth" do
-    assert 1 + 1 == 2
+  defmodule TestRepo do
+    def insert!(record) do
+      send self, {:created, record}
+      record
+    end
+  end
+
+
```
jsteiner: :space_invader:

test/anvil_test.exs
```diff
@@ -1,7 +1,131 @@
...
+  defmodule MyApp.Anvil do
+    use Anvil
+
+    def fixture(:without_opts) do
```
jsteiner: Aren't these more `factories` than `fixtures`?

paulcsmith: I feel like this function is a fixture. The other functions are factory functions to do something with the fixture. Maybe that's wrong though :S

test/anvil_test.exs
paulcsmith: Good catch

lib/anvil.ex
```diff
@@ -1,2 +1,63 @@
...
+      message = "No fixture defined for #{inspect fixture_name}"
+      %UndefinedFixture{message: message}
+    end
+  end
+
+  defmacro __using__(_opts) do
+    quote do
+      @before_compile unquote(__MODULE__)
+
+      def assoc(opts, fixture_name) do
```
BlakeWilliams: Thoughts on putting these methods outsude if `__using__` or a separate module so you can document them? https://github.com/BlakeWilliams/Elixir-Slack/blob/master/lib/slack.ex#L75

test/anvil_test.exs
BlakeWilliams: Extra newline

test/anvil_test.exs
paulcsmith: Actually, maybe factory makes the most sense here. I'll think about it a bit more

paulcsmith: @BlakeWilliams Any thoughts on `factory` versus `fixture` here?

lib/anvil.ex
paulcsmith: I think that is a great idea, but I think the problem is that it doesn't actually define the functions. It only import them. We need to define the functions so you can call `create`, `build`, etc. 

Instead I can define the function in the using macro and then call function in `Anvil`. I think that is a good balance between the two

PR #6, opened 2015-09-15
Add example using the module in a test

paulcsmith:


lib/ex_machina.ex
```diff
@@ -4,6 +4,8 @@ defmodule ExMachina do
 
   ## Examples
 
+  Put this somewhere like `test/support/ex_machina.ex`
```
BlakeWilliams: `Create a file in your application for ex machina: e.g. `test/support/ex_machina.ex``?

paulcsmith: Not sure about that. It is more specific, but longer too. What did you dislike about "Put this somewhere like `test/support/ex_machina.ex`?

paulcsmith: Maybe `Create a module for your factories. Often somewhere like `test/support/machina.ex`?

BlakeWilliams: I like, that, thoughts on `e.g.` instead of `Often somewhere like`?

BlakeWilliams:
LGTM!

PR #8, opened 2015-09-15
Allow specifying a different factory name for associations

jsteiner:


lib/ex_machina.ex
```diff
@@ -65,26 +65,36 @@ defmodule ExMachina do
...
-    else
-      ExMachina.create(module, factory_name)
+  def assoc(module, attrs, factory_name, opts \\ []) do
+    case Map.get(attrs, factory_name) do
+      nil -> create_assoc(module, factory_name, opts)
+      record -> record
     end
   end
 
+  defp create_assoc(module, _factory_name, [factory: factory_name]) do
```
paulcsmith: Super nitpicky, but you can remove the `[]` in the pattern match if you want

test/ex_machina_test.exs
```diff
@@ -85,8 +83,18 @@ defmodule ExMachinaTest do
     assert comment.article == my_article
   end
 
-  test "factorys can be defined without the second options param" do
-    assert MyApp.ExMachina.build(:without_opts) == MyApp.ExMachina.factory(:without_opts)
```
paulcsmith: Did you mean to delete this test?

test/ex_machina_test.exs
```diff
@@ -85,8 +83,18 @@ defmodule ExMachinaTest do
...
+  test "assoc/3 can specify a factory for the association" do
```
paulcsmith: I think it would be good to put this above line 78. So we test the method directly, and then at the end test that it can be used inside the module

test/ex_machina_test.exs
```diff
@@ -85,8 +83,18 @@ defmodule ExMachinaTest do
...
+    attrs = %{}
+
+    account = ExMachina.assoc(MyApp.ExMachina, attrs, :account, factory: :user)
+
+    newly_created_account = %{id: 3, admin: false, name: "John Doe"}
+    assert account == newly_created_account
+    assert_received {:created, ^newly_created_account}
+  end
+
+  test "factorys can be defined without the attrs param" do
```
paulcsmith: factor*ies* :)

PR #9, opened 2015-09-15
Implement fields_for

jsteiner:


lib/ex_machina.ex
```diff
@@ -96,6 +100,32 @@ defmodule ExMachina do
...
+
+      # Returns %{name: "John Doe", admin: true}
+      fields_for(:user, admin: true)
+  """
+  def fields_for(module, factory_name, attrs \\ %{}) do
+    module.build(factory_name, attrs)
+    |> drop_undesireable_fields
+  end
+
+  defp drop_undesireable_fields(record = %{__struct__: struct}) do
```
paulcsmith: This is pretty Ecto specific. Should this be called `drop_unwanted_ecto_fields`?

lib/ex_machina.ex
```diff
@@ -96,6 +100,32 @@ defmodule ExMachina do
...
+    record
+    |> Map.drop([:__struct__, :__meta__])
+    |> Map.drop(struct.__schema__(:associations))
+  end
+  defp drop_undesireable_fields(record) do
+    record
```
paulcsmith: I would consider raising an error here. If they pass in a non Ecto struct, then they're probably using it wrong. Maybe something like `fields_for expects an Ecto model. Instead it was called on #{inspect record}`

lib/ex_machina.ex
```diff
@@ -96,6 +100,32 @@ defmodule ExMachina do
   end
 
   @doc """
+  Builds a factory with the passed in factory_name and returns its fields
+
```
paulcsmith: I think it would be good to mention that it is expecting an Ecto struct, and that it will raise an error if you try to build something other than that.

lib/ex_machina.ex
BlakeWilliams: Would `http://elixir-lang.org/docs/stable/elixir/` help here at all?

BlakeWilliams: Oops, docs don't link properly. https://github.com/elixir-lang/elixir/blob/v1.0.5/lib/elixir/lib/map.ex#L60

lib/ex_machina.ex
```diff
@@ -96,6 +100,32 @@ defmodule ExMachina do
...
+      fields_for(:user, admin: true)
+  """
+  def fields_for(module, factory_name, attrs \\ %{}) do
+    module.build(factory_name, attrs)
+    |> drop_undesireable_fields
+  end
+
+  defp drop_undesireable_fields(record = %{__struct__: struct}) do
+    record
+    |> Map.drop([:__struct__, :__meta__])
```
paulcsmith: What do you think of?

```
|> Map.from_struct
|> Map.delete(:__meta__)
```

Though I'd be fine leaving it as-is too. Not sure which is more clear, just throwing ideas out there

lib/ex_machina.ex
paulcsmith: I'm leaning toward no. If someone just wants a map, they can use `Map.from_struct`. I think this function is supposed to be used for Ecto records, but I see your point. Maybe we try raising the error and see if we actually need/want fields_for to convert to a struct if it's not an Ecto struct?

test/ex_machina_test.exs
```diff
@@ -103,6 +117,21 @@ defmodule ExMachinaTest do
...
 
+  test "fields_for/2 returns the fields for the matching factory" do
+    assert MyApp.ExMachina.fields_for(:user) == %{
+      id: 3,
+      admin: false,
+      name: "John Doe"
+    }
+  end
+
+  test "fields_for/2 removes undesireable fields from structs" do
```
paulcsmith: `undeseriable Ecto fields` to make it clear that's specific to Ecto in this case?

lib/ex_machina.ex
BlakeWilliams: I'd assume `fields_for` would return a plain map with the fields, is that not the case here?

paulcsmith:
Done commenting. I think this should be really helpful once it's in :D

lib/ex_machina.ex
BlakeWilliams: Oh I see, I think it should raise if you just hand it a record.

paulcsmith: Yeah that's a good point actually...so yeah maybe that's the way to think about it. It strips Ecto fields if it's an Ecto Struct and returns a plain map, otherwise. I would be done for that. I guess I was just thinking about it weirdly

paulcsmith: So raise if it's a plan map, convert to a plain map if it's a struct?

BlakeWilliams: It's private, so should we even run into that situation?

paulcsmith: Yes, if the factory returns a map

```
def factory(:config), do: %{foo: :bar}
```

lib/ex_machina.ex
```diff
@@ -96,6 +100,35 @@ defmodule ExMachina do
   end
 
   @doc """
+  Builds a factory with the passed in factory_name and returns its fields
+
+  This is only for use with Ecto models.
```
paulcsmith: Maybe add what it will do at the end? `Will return a map with the fields and virtual fields, but without the Ecto metadata and associations.`?

lib/ex_machina.ex
```diff
@@ -96,6 +100,35 @@ defmodule ExMachina do
...
+  end
+
+  defp drop_ecto_fields(record = %{__struct__: struct, __meta__: %{__struct__: Ecto.Schema.Metadata}}) do
+    record
+    |> Map.from_struct
+    |> Map.delete(:__meta__)
+    |> Map.drop(struct.__schema__(:associations))
+  end
+  defp drop_ecto_fields(record) do
+    raise ArgumentError, "#{inspect record} is not an Ecto model. Use `build` instead."
```
paulcsmith: :+1:

paulcsmith:
Just had one comment on the documentation. The rest looks awesome!

PR #11, opened 2015-09-16
Simplify factory definition

paulcsmith:
@jsteiner @BlakeWilliams I think this is nicer to read and to write. It also makes it so we don't get warning about grouping similar functions together.

The only downside I see is that you have to define `_attrs`, but I still think that's cleaner than what we had before

What do you think?


paulcsmith:
Closing this because we're working on a "better" way to do this. The problem with this solution is that when you import it you now have a bunch of functions that you probably don't want to import

PR #12, opened 2015-09-16
Extract all Ecto functionality to separate module

jsteiner:
This should make it easier to add other adapters in the future and
reduces the need for conditionals due to separation of concerns.

### DIFF
--- a/lib/ex_machina.ex
+++ b/lib/ex_machina.ex
@@ -20,16 +20,15 @@ defmodule ExMachina do
 
   defmodule UndefinedSave do
     @moduledoc """
-    Error raised when trying to call create and no repo or save_function is
-    defined.
+    Error raised when trying to call create and save_record/1 is
+    not defined.
     """
 
     defexception [:message]
 
     def exception do
       %UndefinedSave{
-        message: "Define save_function/1 or include the repo option. See docs
-        for ExMachina.save_record."
+        message: "Define save_record/1. See docs for ExMachina.save_record/1."
       }
     end
   end
@@ -38,20 +37,11 @@ defmodule ExMachina do
 
   def start(_type, _args), do: ExMachina.Sequence.start_link
 
-  defmacro __using__(opts) do
+  defmacro __using__(_opts) do
     quote do
       @before_compile unquote(__MODULE__)
-      @repo Dict.get(unquote(opts), :repo)
 
       import ExMachina, only: [sequence: 2]
-
-      defp assoc(attrs, factory_name, opts \\ []) do
-        ExMachina.assoc(__MODULE__, attrs, factory_name, opts)
-      end
-
-      def fields_for(factory_name, attrs \\ %{}) do
-        ExMachina.fields_for(__MODULE__, factory_name, attrs)
-      end
 
       def build(factory_name, attrs \\ %{}) do
         ExMachina.build(__MODULE__, factory_name, attrs)
@@ -67,10 +57,6 @@ defmodule ExMachina do
 
       def create_list(number_of_factories, factory_name, attrs \\ %{}) do
         ExMachina.create_list(__MODULE__, number_of_factories, factory_name, attrs)
-      end
-
-      def save_record(record) do
-        ExMachina.save_record(__MODULE__, @repo, record)
       end
     end
   end
@@ -88,69 +74,6 @@ defmodule ExMachina do
       end
   """
   def sequence(name, formatter), do: ExMachina.Sequence.next(name, formatter)
-
-  @doc """
-  Gets a factory from the passed in attrs, or creates if none is present
-
-  ## Examples
-
-      attrs = %{user: %{name: "Someone"}}
-      # Returns attrs.user
-      assoc(attrs, :user)
-
-      attrs = %{}
-      # Creates and returns new instance based on :user factory
-      assoc(attrs, :user)
-
-      attrs = %{}
-      # Creates and returns new instance based on :user factory
-      assoc(attrs, :author, factory: :user)
-  """
-  def assoc(module, attrs, factory_name, opts \\ []) do
-    case Map.get(attrs, factory_name) do
-      nil -> create_assoc(module, factory_name, opts)
-      record -> record
-    end
-  end
-
-  defp create_assoc(module, _factory_name, factory: factory_name) do
-    ExMachina.create(module, factory_name)
-  end
-  defp create_assoc(module, factory_name, _opts) do
-    ExMachina.create(module, factory_name)
-  end
-
-  @doc """
-  Builds a factory with the passed in factory_name and returns its fields
-
-  This is only for use with Ecto models.
-
-  Will return a map with the fields and virtual fields, but without the Ecto
-  metadata and associations.
-
-  ## Example
-
-      def factory(:user) do
-        %MyApp.User{name: "John Doe", admin: false}
-      end
-
-      # Returns %{name: "John Doe", admin: true}
-      fields_for(:user, admin: true)
-  """
-  def fields_for(module, factory_name, attrs \\ %{}) do
-    module.build(factory_name, attrs)
-    |> drop_ecto_fields
-  end
-
-  defp drop_ecto_fields(record = %{__struct__: struct, __meta__: %{__struct__: Ecto.Schema.Metadata}}) do
-    record
-    |> Map.from_struct
-    |> Map.delete(:__meta__)
-    |> Map.drop(struct.__schema__(:associations))
-  end
-  defp drop_ecto_fields(record) do
-    raise ArgumentError, "#{inspect record} is not an Ecto model. Use `build` instead."
-  end
 
   @doc """
   Builds a factory with the passed in factory_name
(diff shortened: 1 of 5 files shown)


lib/ex_machina/ecto.ex
```diff
@@ -0,0 +1,58 @@
+defmodule ExMachina.Ecto do
+  defmacro __using__(opts) do
+    quote do
+      use ExMachina
+
+      @repo Dict.get(unquote(opts), :repo)
```
jsteiner: What is the best way to make this option required?

lib/ex_machina.ex
```diff
@@ -20,16 +20,15 @@ defmodule ExMachina do
 
   defmodule UndefinedSave do
     @moduledoc """
-    Error raised when trying to call create and no repo or save_function is
-    defined.
+    Error raised when trying to call create and save_record/1 is
+    not defined.
```
paulcsmith: Heh, nice catch :)

lib/ex_machina/ecto.ex
paulcsmith: Dict.fetch! would blow up if you try to call it without repo defined in the opts. However if you want something more friendly you could probably do something in save record like we did before. So you would check if `@repo` is defined. If it is, save it, if not raise a helpful error letting them know they need to pass `repo` in to the opts

lib/ex_machina.ex
```diff
@@ -279,10 +194,41 @@ defmodule ExMachina do
...
+      If using ExMachina.Ecto (`use ExMachina.Ecto, repo: MyApp.Repo`) this
+      function will call `insert!` on the passed in repo.
+
+      If you are not using ExMachina.Ecto, you must define a custom
+      save_record/1 for saving the record.
+
+      ## Examples
+
+          defmodule MyApp.Factories do
+            use ExMachina, repo: MyApp.Repo
```
paulcsmith: use ExMachina.Ecto

lib/ex_machina.ex
```diff
@@ -279,10 +194,41 @@ defmodule ExMachina do
...
+
+            def factory(:user), do: %User{name: "John"}
+          end
+
+          # Will build and save the record to the MyApp.Repo
+          MyApp.Factories.create(:user)
+
+          defmodule MyApp.JsonFactories do
+            # Note `repo` was not passed as an option
```
paulcsmith: Note we are using just `ExMachina` without Ecto and without the `repo` option?

lib/ex_machina/ecto.ex
jsteiner: Is there any way to get it to crash at compile time but also provide a friendly message? Maybe `Dict.fetch!` is good enough?

lib/ex_machina/ecto.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  defp drop_ecto_fields(record = %{__struct__: struct, __meta__: %{__struct__: Ecto.Schema.Metadata}}) do
+    record
+    |> Map.from_struct
+    |> Map.delete(:__meta__)
+    |> Map.drop(struct.__schema__(:associations))
+  end
+  defp drop_ecto_fields(record) do
+    raise ArgumentError, "#{inspect record} is not an Ecto model. Use `build` instead."
+  end
```
paulcsmith: I'm wondering if `assoc` should be only in Ecto and should return the id automatically. It's only useful for Ecto, because you can just use `build(:association)` for other types of factories. What do you think?

lib/ex_machina/ecto.ex
paulcsmith: If you use Dict.fetch! It should crash at compile time, but I don't know of a way to get a nice error message. You could try get the option before the `quote` and raising there, but I've never done that

test/ex_machina/ecto_test.ex
```diff
@@ -0,0 +1,77 @@
...
+  test "create_list/3 creates factory and saves it passed in number of times" do
+    users = MyApp.ExMachina.create_list(3, :user, admin: true)
+
+    created_user = %{id: 3, name: "John Doe", admin: true}
+    assert users == [created_user, created_user, created_user]
+    assert_received {:created, ^created_user}
+    assert_received {:created, ^created_user}
+    assert_received {:created, ^created_user}
+    refute_received {:created, _}
+  end
```
paulcsmith: I think we could remove all of the `create...` tests and instead call `save_record` and ensure that it saves. This would simplify this test file a lot I think. 

create, create_list, etc. are tested in the other file already

test/ex_machina_test.exs
```diff
@@ -171,14 +131,15 @@ defmodule ExMachinaTest do
     user = MyApp.ExMachina.create(:user, admin: true)
 
     assert user == %{id: 3, name: "John Doe", admin: true}
-    assert_received {:created, %{name: "John Doe", admin: true}}
+    assert_received {:custom_save, %{name: "John Doe", admin: true}}
   end
 
   test "create/2 builds factory and performs custom save if repo is not set" do
```
paulcsmith: "...and performs save with user defined `save_record/1`"

lib/ex_machina/ecto.ex
jsteiner: Sounds good. We can always move it back later if we have to.

test/ex_machina_test.exs
```diff
@@ -171,14 +131,15 @@ defmodule ExMachinaTest do
     user = MyApp.ExMachina.create(:user, admin: true)
 
     assert user == %{id: 3, name: "John Doe", admin: true}
-    assert_received {:created, %{name: "John Doe", admin: true}}
```
paulcsmith: We can probably remove this test and just use the one below since we aren't doing the repo definition here anymore

test/ex_machina_test.exs
```diff
@@ -1,30 +1,9 @@
...
-  defmodule MyApp.Book do
-    defstruct title: nil, publisher: nil, __meta__: %{__struct__: Ecto.Schema.Metadata}, publisher_id: 1
-
-    def __schema__(:associations) do
-      [:publisher]
-    end
-  end
-
   defmodule MyApp.ExMachina do
     use ExMachina, repo: TestRepo
```
paulcsmith: I think this should not have the `repo` option right?

paulcsmith:
Left a couple comments. This look really awesome! I think this was a great idea. Also, we'll need to update the README

lib/ex_machina/ecto.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  @doc """
+  Saves a record using `Repo.insert!` when `create` is called.
+  """
+  def save_record(module, repo, record) do
+    if repo do
+      repo.insert!(record)
```
paulcsmith: I think the simplest thing to do would be to raise a helpful error here if the repo is nil

test/ex_machina_test.exs
```diff
@@ -171,14 +131,15 @@ defmodule ExMachinaTest do
...
+    record = MyApp.ExMachina.create(:user)
 
-    assert record == %{foo: :bar}
-    assert_received {:custom_save, %{foo: :bar}}
+    created_record = %{admin: false, id: 3, name: "John Doe"}
+    assert record == created_record
+    assert_received {:custom_save, ^created_record}
   end
 
   test "create/2 raises a helpful error if no repo and no custom function" do
```
paulcsmith: replace "if no repo and no custom function" with "if no custom save_record/1 function is defined" or something like that?

lib/ex_machina/ecto.ex
```diff
@@ -0,0 +1,93 @@
...
+      use ExMachina
+
+      @repo Dict.fetch!(unquote(opts), :repo)
+
+      def fields_for(factory_name, attrs \\ %{}) do
+        ExMachina.Ecto.fields_for(__MODULE__, factory_name, attrs)
+      end
+
+      defp assoc(attrs, factory_name, opts \\ []) do
+        ExMachina.assoc(__MODULE__, attrs, factory_name, opts)
```
paulcsmith: Does this work? I think it's defined on `ExMachina.Ecto`

test/ex_machina/ecto_test.ex
```diff
@@ -0,0 +1,115 @@
...
+        id: 1,
+        title: "My Awesome Article",
+        author_id: assoc(attrs, :author, factory: :user).id
+      }
+    end
+
+    def factory(:comment, attrs) do
+      %{
+        body: "This is great!",
+        article_id: assoc(attrs, :article).id
```
paulcsmith: Did we want to make this return id by default, or leave it as-is?

test/ex_machina/ecto_test.ex
```diff
@@ -0,0 +1,115 @@
...
+
+  defmodule MyApp.Book do
+    defstruct title: nil, publisher: nil, __meta__: %{__struct__: Ecto.Schema.Metadata}, publisher_id: 1
+
+    def __schema__(:associations) do
+      [:publisher]
+    end
+  end
+
+  defmodule MyApp.ExMachina do
```
paulcsmith: What do you think of calling this just `EctoFactories`?

test/ex_machina/ecto_test.ex
jsteiner: I think we can leave it as is for now. Now that we've extracted it to it's own module, we can make another commit to allow for `article: assoc(attrs, :article)`

paulcsmith: Ah ok, good idea!

paulcsmith:
:+1: looks great!

PR #13, opened 2015-09-17
Fix Ecto tests. Make it easier to do the right thing with `assoc/3`

paulcsmith:
The rationale behind this is that it should be easy to do the right thing, and hard to do the wrong thing. Associations should be set using the name of the association and not the id. 

I believe this helpful for 2 reasons
1) It makes it so you don't accidentally create an associated record without meaning to
2) It makes it clear how you should set associations. I've seen plenty of projects that set factory associations with `user` or `user_id` depending on the dev. This aims to help fix that.

### DIFF
--- a/test/ex_machina/ecto_test.exs
+++ b/test/ex_machina/ecto_test.exs
@@ -105,6 +105,14 @@ defmodule ExMachina.EctoTest do
     assert_received {:created, ^newly_created_account}
   end
 
+  test "assoc/3 raises helpful error if setting the id directly" do
+    message = "Set association with :article instead of :article_id"
+    assert_raise ArgumentError, message, fn ->
+      MyApp.EctoFactories.create(:comment, article_id: 1)
+    end
+    refute_received {:custom_save, _}
+  end
+
   test "can use assoc/3 in a factory to override associations" do
     my_article = MyApp.EctoFactories.create(:article, title: "So Deep")
 
--- a/test/ex_machina/ecto_test.exs
+++ b/test/ex_machina/ecto_test.exs
@@ -71,38 +71,38 @@ defmodule ExMachina.EctoTest do
   end
 
   test "save_record/1 passes the data to @repo.insert!" do
-    user = MyApp.EctoFactories.save_record(:user, admin: true)
+    record = MyApp.EctoFactories.save_record(%{foo: "bar"})
 
-    assert user == %{id: 3, name: "John Doe", admin: true}
-    assert_received {:created, %{name: "John Doe", admin: true}}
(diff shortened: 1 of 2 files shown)


paulcsmith:
Either way I will need to merge the first commit since it fixes the broken Ecto tests

test/ex_machina/ecto_test.exs
```diff
@@ -71,38 +71,46 @@ defmodule ExMachina.EctoTest do
...
+    account = ExMachina.Ecto.assoc(MyApp.EctoFactories, attrs, :account, factory: :user)
 
     newly_created_account = %{id: 3, admin: false, name: "John Doe"}
     assert account == newly_created_account
-    assert_received {:custom_save, ^newly_created_account}
+    assert_received {:created, ^newly_created_account}
+  end
+
+  test "assoc/3 raises helpful error if setting the id directly" do
+    message = "Set association with :article instead of :article_id"
```
BlakeWilliams: I'd rather set with `_id` since it's always available rather than the association directly to be honest.

paulcsmith: Could you give some examples of when you would need to set the _id and not use a whole record? I'm curious what the use cases are. It may be that I need to rethink how we do associations

The problem with using `_id` is that if you are using `assoc` in your factory, it will create a record. Which is likely not what you expect

BlakeWilliams: Maybe that should do some magic and just pull out the `id`? One case I can think of is records that haven't preloaded associations.

BlakeWilliams: I think it would make sense not to return the full association by default, since that's not how the record comes by default anyways which would force `Repo.preload`.

paulcsmith: @BlakeWilliams Yeah that could be the case. I think I'll add this just to Constable and see how it works out. That way we don't pollute this Repo with experiments :)

Could you explain this more though:

> I think it would make sense not to return the full association by default, since that's not how the record comes by default anyways which would force Repo.preload

Would it just return the id then? I'm not sure I understand

BlakeWilliams: Yeah,  I'd expect it to return just the ID like the Ecto methods do.

paulcsmith: Ok I see. I do agree with that, but I think we might want a different name than `assoc` if we do that. Something like `get_id(attrs, :user)` or `assoc_id(attrs, :user)`. Thoughts on that?

paulcsmith:
Closing this. The Ecto tests are passing, but we decided not to throw the error

Issue #14, opened 2015-09-22
Pattern for handling `has_one` and `has_many` associations


jesenko:
Is there a way to handle creation of `has_one` and `has_many` associations? Namely, to have something like:

```elixir
def factory(:article) do
  %Article{
    title: "My Awesome Article",
    comments: %Comment{
      body: "I will be always here!"
    }
  }
end
```
and to create it 
```elixir
article_with_comment = create(:article)
```

Current, if I try to do this I get error  `... Assocs can only be manipulated via changesets, be it on insert, update or delete.` when `Repo.insert!` [is called](https://github.com/thoughtbot/ex_machina/blob/master/lib/ex_machina/ecto.ex#L90).

If instead model to insert is converted to changeset first, associations are inserted as well,
```elixir
%{__struct__: model} = record
struct(model) |> Ecto.Changeset.change(drop_ecto_fields(record)) |> repo.insert!
```

Not sure if this is the right approach though...

paulcsmith:
Thanks for this issue and the PR, another way to handle this if you want to make things a bit more flexible is to use functions. Here's something we've been doing lately.

```elixir
def with_comment(article) do
  create(:comment, article: article)
  article
end

create(:article) |> with_comment
```

That way you can add comments only when you need them. It works great for creating join tables as well. I still agree we need a way to handle `has_one` and `has_many` associations though for cases where this isn't ideal

jsteiner:
Closing this since #15 was merged into 0.4.0.

PR #15, opened 2015-09-22
Handle has_one and has_many associations

jesenko:
Ecto.Repo.insert! only handles associations when changeset is passed as
argument. Prior to insertion, Ecto.Model is thus converted to
Ecto.Changeset. Related to #14.

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -103,17 +103,55 @@ defmodule ExMachina.Ecto do
     ExMachina.build(module, factory_name)
   end
 
+  defp get_assocs(%{__struct__: struct}) do
+    for a <- struct.__schema__(:associations) do
+      {a, struct.__schema__(:association, a)}
+    end
+  end
+
+  defp belongs_to_assocs(model) do
+    for {a, %Ecto.Association.BelongsTo{}} <- get_assocs(model), do: a
+  end
+
+  defp not_loaded_assocs(model) do
+    for {a, %Ecto.Association.Has{}} <- get_assocs(model),
+      !Ecto.Association.loaded?(Map.get(model, a)),
+      do: a
+  end
+
+  defp reset_associations(target, source) do
+    target
+      |> belongs_to_assocs
+      |> Enum.reduce(target, fn(a, target) -> Map.put(target, a, Map.get(source, a)) end)
+  end
+
+  defp convert_to_changes(record) do
+    record
+    |> Map.from_struct
+    |> Map.delete(:__meta__)
+    # drop fields for `belongs_to` assocs as they cannot be handled by changeset
+    |> Map.drop(belongs_to_assocs(record))
+    |> Map.drop(not_loaded_assocs(record))
+  end
+
   @doc """
   Saves a record and all associated records using `Repo.insert!`
   """
-  def save_record(module, repo, record) do
-    record
-    |> associate_records(module)
+  def save_record(module, repo, %{__struct__: model} = record) do
+    record = record |> associate_records(module)
+    changes = record |> convert_to_changes
+
+    struct(model)
+    |> Ecto.Changeset.change(changes)
     |> repo.insert!
+    |> reset_associations(record)
+  end
+  def save_record(_, _ , record) do
+    raise ArgumentError, "#{inspect record} is not Ecto model."
   end
 
-  defp associate_records(built_record = %{__struct__: struct}, module) do
-    association_names = struct.__schema__(:associations)
+  defp associate_records(built_record, module) do
+    association_names = belongs_to_assocs(built_record)
 
     Enum.reduce association_names, built_record, fn(association_name, record) ->
       case association = Map.get(record, association_name) do
--- a/test/ex_machina/ecto_test.exs
+++ b/test/ex_machina/ecto_test.exs
@@ -1,5 +1,5 @@
 defmodule ExMachina.EctoTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
   alias ExMachina.TestRepo
 
   setup_all do
@@ -144,6 +144,12 @@ defmodule ExMachina.EctoTest do
 
     assert_raise ArgumentError, message, fn ->
       Factory.save_record(%Article{title: "Ecto is Awesome", author: author})
+    end
+  end
+
+  test "save_record/1 raises unless Ecto.Model is passed" do
+    assert_raise ArgumentError, ~r"not Ecto model", fn ->
+      Factory.save_record(%{foo: "bar"})
     end
   end
 
--- a/priv/test_repo/migrations/20151016110321_create_packages_and_invoices.exs
+++ b/priv/test_repo/migrations/20151016110321_create_packages_and_invoices.exs
@@ -1,0 +1,17 @@
+defmodule ExMachina.TestRepo.Migrations.CreatePackagesAndInvoices do
+  use Ecto.Migration
+
+  def change do
+    create table(:invoices) do
+      add :title, :string
+      add :package_id, :integer
+    end
+    create table(:package_statuses) do
+      add :status, :string
+      add :package_id, :integer
+    end
+    create table(:packages) do
+      add :description, :string
+    end
+  end
+end
--- a/test/ex_machina/ecto_has_many_test.exs
+++ b/test/ex_machina/ecto_has_many_test.exs
@@ -1,0 +1,105 @@
+defmodule ExMachina.EctoHasManyTest do
+  use ExUnit.Case, async: false
+  alias ExMachina.TestRepo
+
+  setup_all do
+    Ecto.Adapters.SQL.begin_test_transaction(TestRepo, [])
+    on_exit fn -> Ecto.Adapters.SQL.rollback_test_transaction(TestRepo, []) end
+    :ok
+  end
+
+  setup do
+    Ecto.Adapters.SQL.restart_test_transaction(TestRepo, [])
+    :ok
+  end
+
+  defmodule Package do
+    use Ecto.Model
+    schema "packages" do
+      field :description, :string
+      has_many :statuses, ExMachina.EctoHasManyTest.PackageStatus
+    end
+  end
+
+  defmodule PackageStatus do
+    use Ecto.Model
+    schema "package_statuses" do
+      field :status, :string
+      belongs_to :package, Package
+    end
+  end
+
+  defmodule Invoice do
+    use Ecto.Model
+    schema "invoices" do
+      field :title, :string
+      belongs_to :package, Package
+    end
+  end
+
+  defmodule Factory do
+    use ExMachina.Ecto, repo: TestRepo
+
+    factory(:invalid_package) do
+      %Package{
+        description: "Invalid package without any statuses"
+      }
+    end
+
+    factory(:package) do
+      %Package{
+        description: "Package that just got ordered",
+        statuses: [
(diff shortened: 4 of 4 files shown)


paulcsmith:
@jesenko Thanks for the issue and the PR! Handling `has_one` and `has_many` associations is definitely a priority. I think you're right that using Ecto Changesets is the way to handle this. There are a couple of things we need to figure out before merging this in though.

1) What happens if they are using ExMachina and don't have Ecto as a dependency? Right now I think it would fail to compile. I'm not sure what the best way to handle this is. We could extract `ExMachina.Ecto` but I would rather not do that quite yet.
2) This would need a couple tests to handle various cases such as `create(:article, comments: build_pair(:comments))` and using it from within a the factory definition like in #14. If you need help with the test, let me know! I'm happy to help.

jesenko:
Great! I was not sure I was on the right track with Ecto.Changeset...

Maybe instead of using `Ecto.Changeset`, we could manually transform `Model` struct into `Changeset` struct prior to sending it to `Repo.insert`? This would eliminate dependency on Ecto; In this case, tests could be added that would verify whether appropriate changeset struct is passed to `Repo.insert!` Not sure yet how involved Model -> Changeset transformation is, will look into it...

If we instead use `Ecto.Changeset` explicitly, I am not sure how to approach tests. If input to `Repo.insert!` would be verified, we could only check whether it was transform into changeset - however for tests to pass, we would need explicit dep to Ecto again...

paulcsmith:
@jesenko I think it's ok to add Ecto as a development dependency at this point, I just don't want to make it a runtime dependency. People should be able to use the library without Ecto. So we can use `Code.ensure_loaded?` to see if it's loaded or not, if it is, then it can be run through the changes function.

Not sure if that makes sense. This might be a bit complicated. I can take a crack at it this Friday if you need some help to get started

jesenko:
@paulcsmith Agreed, might be best to stick with `Ecto.Changeset`. However, I am not sure about why and how to use `Code.ensure_loaded?` - is user will use `ExMachina` instead of `ExMachina.Ecto`, no runtime error should appear as Ecto.Changeset is used only within ExMachina.Ecto module. If using `ExMachina.Ecto` however, runtime error will appear whenever `create` and related functions will be called. I think that error is appropriate in this case (not sure why someone would be using `ExMachina.Ecto` without Ecto). Maybe, more user friendly error message could be provided using `Code.ensure_loaded?`, notifying user to add `ecto` as deps.

As far as tests are concerned, I would rather wait for #13 to be merged, in master tests related to `Ecto` are failing. I'll add them to PR then...

paulcsmith:
@jesenko The tests are passing on master. That was closed by #16 in a separate PR. Sorry for the confusion.

I think that it should work like this

* If they are not using `ExMachina.Ecto` everything should work as expected. No errors.
* If they try to `use ExMachina.Ecto` a helpful error should be raised letting them know that they need to add `ecto` to their dependencies if `Code.ensure_loaded?` fails. Something like `"You tried to use ExMachina.Ecto, but the Ecto module is not loaded. Please add ecto to your dependencies"`. This could probably done in the `__using__` block of `ExMachina.Ecto`. So if you use it, and it's not loaded, you get the error

What do you think?

jesenko:
@paulcsmith Sorry, missed #16...will rebase PR to latest master.

:+1: on the handling Ecto dependency

paulcsmith:
Another thing to note is that `drop_ecto_fields` also drops the association fields. So a similar function that doesn't drop the association fields will need to be written.

jesenko:
I updated PR with some refinements in behaviour and additional tests. I also introduced Ecto.Models into tests (instead of ordinary maps / structs), as now Ecto is dev dependency...

lib/ex_machina/ecto.ex
```diff
@@ -82,12 +91,32 @@ defmodule ExMachina.Ecto do
...
   Saves a record using `Repo.insert!` when `create` is called.
   """
-  def save_record(module, repo, record) do
-    if repo do
-      repo.insert!(record)
-    end
+  def save_record(module, repo, %{__struct__: model} = record) do
+    struct(model)
+    # `belongs_to` assocs are not managed by changeset, add them to model
+    |> Map.merge(Map.take(record, belongs_to_assocs(record)))
```
paulcsmith: What happens if they are included? Is an error thrown?

test/ex_machina/ecto_test.exs
```diff
@@ -111,5 +171,34 @@ defmodule ExMachina.EctoTest do
...
+    end
+  end
+
+  test "create/2 creates model with `has_many` associations" do
+    article = MyApp.EctoFactories.create(:article_with_comments)
+
+    assert %{comments: [%{body: "This is great!"}]} = article
+    assert_received {:created, %{model: %MyApp.Article{}} = article_changeset}
+    assert %{changes: %{comments: [%{action: :insert}]}} = article_changeset
+    refute_received {:created, %{model: %MyApp.Comment{}}}
```
jsteiner: Don't we want this to be created?

lib/ex_machina/ecto.ex
jesenko: @paulcsmith `belongs_to` associations can be included in factory or `attrs`, they will not be included in changeset changes, however they will be retained in the resulting model.

For example, this works:
```elixir
user = create(:user)
article = create(:article, author: user)
IO.inspect article.author.name # author is retained in `create`
```
Ecto.Changeset raises error on insert if `belongs_to` association is listed in `%Ecto.Changeset{changes: %{}}`. They are however already persisted by `assoc` method.

The logic in `save_record` can be summarized as
1) Add all `belongs_to` fields to initial model
2) Create changeset from initial model, NOT changing any `belongs_to` fields
3) Insert changeset

If step 1) is skipped, example above does not work, e.g.
```elixir
IO.inspect article.author.name # raises assoc not loaded exception
```

Does this answer your question?

paulcsmith: Yes that does. Thanks for the explanation :)

test/ex_machina/ecto_test.exs
```diff
@@ -33,19 +73,30 @@ defmodule ExMachina.EctoTest do
...
 
+    # use `inverse: true` to avoid creating article assoc when using comment
+    # factory in `has_many` relation
     def factory(:comment, attrs) do
-      %{
+      comment = %MyApp.Comment{
         body: "This is great!",
-        article_id: assoc(attrs, :article).id
       }
+      case attrs[:inverse] do
```
paulcsmith: I hadn't thought of this case before. This could make the API a lot more complicated for people to use. If we can't rethink this to not use extra attrs, we may not want to add this functionality

test/ex_machina/ecto_test.exs
jesenko: Comments are already created with `article_changeset` (this is handled by Ecto). Individual comments should not be passed to `Repo.insert!` again - this would lead to duplicated entries.

jsteiner:
@jesenko Thanks a lot for this PR! I discussed with @paulcsmith, and we think it's a bit too early for this change. We're still trying to keep the interface small while we figure out how we want this thing to be used. At the moment, we still haven't struggled with this problem. We've mostly been creating associations through the `belongs_to` which still works great.

So, instead of associating comments with an article, we're doing:

```
article = create(:article)
comment = create(:comment, article: article)
```

If you have to create multiple, you can do:

```
article = create(:article)
comments = create_list(:comment, article: article)
```

That being said, we really like your contribution of `verify_ecto_dep`. Would you mind creating a separate PR with that change? We're targeting the `0.3.0` branch at the moment.

Thanks again!

paulcsmith:
As @jsteiner said, thanks for putting this together. We just realized it's a bit too tricky at the moment. Another pattern we've been using (taken from #14)

```
# Defined inside your factory module or anywhere really
def with_comment(article) do
  create(:comment, article: article)
  article
end

# Then you can do.
create(:article) |> with_comment
```

This is nice because you can reuse them easily, pass options if needed and use multiple functions if you want.

`build(:article) |> published_yesterday |> with_comment`

test/ex_machina/ecto_test.exs
jesenko: I just used this as a way to use `create(:comment) and `comments: `[build(:comment)]` with same factory; when used as `comments: [build(:comment)]`, `assoc` should not be invoked as it creates it's own instance of `article`, instead of setting it's parent. Probably, separate factories should be used for these two cases, e.g. `comment_with_article` and `comment`.

test/ex_machina/ecto_test.exs
jsteiner: Ahh, got it!

jesenko:
@paulcsmith @jsteiner OK, will create separate PR for `verify_ecto_dep`. I agree that this change can wait for usage patterns to establish...

Maybe some background on why I needed/wanted this change. We have some models that are coupled very closely to its children, e.g. `%Message{events: [%Events{timestamp, event}]}` where in events, all lifetime events of `Message` are listed. Thus each new Message should already have `%Event{timestamp, event: created}`. Currently, we would always have to use `create(:message) |> with_created_event` which is a bit verbose. Also, we cannot use `assoc(:message)` in other factories as it results in _invalid_ `message`.

What about `Embeds`? According to [this](http://blog.plataformatec.com.br/2015/08/working-with-ecto-associations-and-embeds/) they also can be saved only via changeset?

paulcsmith:
@jesenko Thanks for being so understanding. I really appreciate the example use case. I wonder if you could paste the factory module and the test case so it would be a bit easier to see what's going on? Maybe the models too? That would help get an idea of what the problem is and how it could be solved

The point about Embeds is valid. We haven't used that yet. Maybe we can start a separate issue for that

jesenko:
Just started rebasing PR for `verify_ecto_dep`... I noticed however that Ecto is not required as `test` or `dev` dependency in current master as Ecto.Model is not explicitly used in any test...Should I create a PR nonetheless?

Will provide factory+test case for the example above asap..

jsteiner:
@jesenko I think it's probably best to switch to using `Ecto.Model` instead of our fake.

jesenko:
@paulcsmith @jsteiner I updated PR with `ecto_has_many_test.exs` which defines its own models to demonstrate my use case better. Domain model is quite simple, `Package`,`PackageStatus` and `Invoice`, with following constraints: 
* `Package` should be always have at least _ordered_ status. 
* `Invoice` should be issued only for package with _shipped_ status.

This can be done quite easily if `has_many` assocs are supported as demonstrated in test cases. If `has_many` is not supported, factories can not be used directly but must always be augmented by some auxiliary methods...

paulcsmith:
@jesenko Thanks for your work on this. We're debating how to go about doing this, but what you did was a big push in the right direction. Thank you!

jesenko:
@jesenko I am glad to hear that :) I will probably rebase this PR to latest master later today - I need to bring my fork up-to-date with upstream anyways...

jesenko:
@paulcsmith Rebased to master. I guess this should also work form embeds...

paulcsmith:
@jesenko Thanks! I think @jsteiner is working on something as well. I'm not sure if we'll go with this PR or a version of it that we've tweaked a bit, but we'll keep you posted. We're hoping to have something soon :) 

And yeah I think you're right that it should handle embedded documents! That would be a huge win

paulcsmith:
Merged into `0.4.0` branch https://github.com/thoughtbot/ex_machina/commit/1ff4198488caa8225563ec2d4262a6f42d7d29be

Thanks for your work @jesenko

PR #17, opened 2015-09-22
Add build_list/3 and build_pair/2

jsteiner:


lib/ex_machina.ex
```diff
@@ -119,7 +153,7 @@ defmodule ExMachina do
 
   ## Example
 
-      # Returns a list of 2 users
+      # Returns a list of 2 persisted users
```
paulcsmith: How about `saved` so it matches the `save_record` function?

test/ex_machina_test.exs
```diff
@@ -69,6 +69,30 @@ defmodule ExMachinaTest do
...
 
+  test "build_pair/2 builds 2 factories" do
+    records = MyApp.Factories.build_pair(:user, admin: true)
+
+    expected_record = %{
+      id: 3,
+      name: "John Doe",
+      admin: true
+    }
+
```
paulcsmith: I think this newline can be removed since it feels like it's part of the assertion phase. Same for the test below. What do you think?

paulcsmith:
Just a couple comments. this look good though.

paulcsmith:
Oh and should this be merged to a 0.3.0 branch instead of master so we can keep all the changes separate until we're ready to merge?

jsteiner:
Merged into the new `0.3.0` branch.

PR #18, opened 2015-09-22
Update README to use `save_record`

paulcsmith:
We changed this to be `save_record`, but forgot to update the README


jsteiner:
:+1:

PR #19, opened 2015-09-22
Move factory definitions to a macro

jsteiner:


lib/ex_machina.ex
```diff
@@ -58,6 +58,17 @@ defmodule ExMachina do
       def create_list(number_of_factories, factory_name, attrs \\ %{}) do
         ExMachina.create_list(__MODULE__, number_of_factories, factory_name, attrs)
       end
+
+      def factory(factory_name, attrs \\ [])
```
paulcsmith: Is this needed? I thought we already had a catch all defined in `before_compile`?

paulcsmith:
This looks really good. I think there are some places in the readme that also need to be updated. This should probably be merged into a `0.3.0` branch as well.

test/ex_machina_test.exs
```diff
@@ -27,7 +27,9 @@ defmodule ExMachinaTest do
   defmodule MyApp.NoSaveFunction do
     use ExMachina
 
-    def factory(:foo), do: %{foo: :bar}
+    factory(:foo) do
```
paulcsmith: I think I prefer the one-liner, but if you don't like it then that's ok. We can keep it

jsteiner: It gives a syntax error now that it's a macro.

paulcsmith: Ah, never mind then :P

lib/ex_machina.ex
jsteiner: It's necessary (with the current implementation) because we are defining two function heads and one of them has defaults.

paulcsmith: I'm a bit confused what are the two function heads? I thought since we have the macro it'll always be `factory(name, attrs)`?

paulcsmith: Oh I see. I think we should just remove the ability to define a function without the `attrs` param. So we can delete the test on line 41 of `ex_machina_test.exs`. Since we have the macro there will always be a factory/2 function

jsteiner: Totally agree. Done.

paulcsmith:
Looks good :+1:

jsteiner:
Merged into 0.3.0

PR #20, opened 2015-09-23
Add Ecto dep to ExMachina.Ecto

jesenko:
This PR adds explicit dependency on Ecto in dev and test envs. It also raises error is user of the library tries to `use ExMachina.Ecto` while not having `Ecto` module.

As Ecto is added as test dependence, tests were refactored to use Ecto.Model instead of fakes.

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -1,5 +1,7 @@
 defmodule ExMachina.Ecto do
   defmacro __using__(opts) do
+    verify_ecto_dep
+
     quote do
       use ExMachina
 
@@ -16,6 +18,13 @@ defmodule ExMachina.Ecto do
       def save_record(record) do
         ExMachina.Ecto.save_record(__MODULE__, @repo, record)
       end
+    end
+  end
+
+  defp verify_ecto_dep do
+    unless Code.ensure_loaded?(Ecto) do
+      raise "You tried to use ExMachina.Ecto, but the Ecto module is not loaded. " <>
+        "Please add ecto to your dependencies."
     end
   end
 
@@ -85,7 +94,7 @@ defmodule ExMachina.Ecto do
   @doc """
   Saves a record using `Repo.insert!` when `create` is called.
   """
-  def save_record(module, repo, record) do
+  def save_record(_module, repo, record) do
     if repo do
       repo.insert!(record)
     end
--- a/mix.exs
+++ b/mix.exs
@@ -30,7 +30,8 @@ defmodule ExMachina.Mixfile do
   defp deps do
     [
       {:ex_doc, "~> 0.9", only: :dev},
-      {:earmark, ">= 0.0.0", only: :dev}
+      {:earmark, ">= 0.0.0", only: :dev},
+      {:ecto, "~> 1.0", only: :test}
     ]
   end
 
--- a/test/ex_machina/ecto_test.exs
+++ b/test/ex_machina/ecto_test.exs
@@ -2,17 +2,51 @@ defmodule ExMachina.EctoTest do
   use ExUnit.Case, async: true
 
   defmodule TestRepo do
-    def insert!(record) do
-      send self, {:created, record}
-      record
+    def insert!(model) do
+      # simulate saving of model by adding id
+      model = %{model | id: 1}
+      send self, {:created, model}
+      model
     end
   end
 
   defmodule MyApp.Book do
-    defstruct title: nil, publisher: nil, __meta__: %{__struct__: Ecto.Schema.Metadata}, publisher_id: 1
+    use Ecto.Model
+    schema "books" do
+      field :title, :string
+      belongs_to :publisher, MyApp.Publisher
+    end
+  end
 
-    def __schema__(:associations) do
-      [:publisher]
+  defmodule MyApp.Publisher do
+    use Ecto.Model
+    schema "publishers" do
+      field :title, :string
+    end
+  end
+
+  defmodule MyApp.User do
+    use Ecto.Model
+    schema "users" do
+      field :name, :string
+      field :admin, :boolean
+    end
+  end
+
+  defmodule MyApp.Article do
+    use Ecto.Model
+    schema "articles" do
+      field :title, :string
+      belongs_to :author, MyApp.User
+      has_many :comments, MyApp.Comment
+    end
+  end
+
+  defmodule MyApp.Comment do
+    use Ecto.Model
+    schema "comments" do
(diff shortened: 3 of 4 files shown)


lib/ex_machina/ecto.ex
```diff
@@ -85,7 +94,7 @@ defmodule ExMachina.Ecto do
   @doc """
   Saves a record using `Repo.insert!` when `create` is called.
   """
-  def save_record(module, repo, record) do
+  def save_record(_module, repo, record) do
```
paulcsmith: This makes me wonder if we should even pass the module to save_record at all. I'll change that in a separate PR

paulcsmith:
This is looking good. I don't have time to review it all, but this is a good step. I'd like to test this on a project with and without Ecto as well, just to make sure it's working as expected. I'm not sure if it's possible to test with `Code.ensure_loaded` so testing manually seems like the only way to go

mix.exs
```diff
@@ -30,7 +30,8 @@ defmodule ExMachina.Mixfile do
   defp deps do
     [
       {:ex_doc, "~> 0.9", only: :dev},
-      {:earmark, ">= 0.0.0", only: :dev}
+      {:earmark, ">= 0.0.0", only: :dev},
+      {:ecto, "~> 1.0", only: [:dev, :test]}
```
jsteiner: Should this be only in `:test`?

jesenko: Probably...I think I added it to `dev` to be able to inspect and use it from `IEx` when developing, actually never needed it then...

test/ex_machina/ecto_test.exs
```diff
@@ -100,9 +142,10 @@ defmodule ExMachina.EctoTest do
 
     account = ExMachina.Ecto.assoc(MyApp.EctoFactories, attrs, :account, factory: :user)
 
-    newly_created_account = %{id: 3, admin: false, name: "John Doe"}
+    vals = %{admin: false, name: "John Doe"}
```
paulcsmith: We are typically calling these `attrs`, so I think it would be more clear if it were changed to `attrs` instead of vals

test/ex_machina/ecto_test.exs
```diff
@@ -100,9 +142,10 @@ defmodule ExMachina.EctoTest do
...
+    newly_created_account = struct(MyApp.User, vals) |> Map.put(:id, 1)
```
paulcsmith: Actually I think it would make sense to just create a struct and remove the `vars` variable altogether on this test and the one above it

```
newly_created_user = %MyApp.User{id: 1, name: "John Doe", admin: false}
```

What do you think?

test/ex_machina/ecto_test.exs
```diff
@@ -90,9 +131,10 @@ defmodule ExMachina.EctoTest do
 
     user = ExMachina.Ecto.assoc(MyApp.EctoFactories, attrs, :user)
 
-    created_user = %{id: 3, name: "John Doe", admin: false}
+    vals = %{name: "John Doe", admin: false}
+    created_user = struct(MyApp.User, vals) |> Map.put(:id, 1)
     assert user == created_user
-    assert_received {:created, ^created_user}
+    assert_received {:created, %MyApp.User{}}
```
paulcsmith: I think you could match against the created_user `{:created, ^created_user}` in this test and the one below

test/ex_machina/ecto_test.exs
```diff
@@ -2,17 +2,50 @@ defmodule ExMachina.EctoTest do
   use ExUnit.Case, async: true
 
   defmodule TestRepo do
-    def insert!(record) do
-      send self, {:created, record}
-      record
+    def insert!(model) do
+      send self, {:created, model}
+      # simulate saving of model by adding id
+      model |> Map.put(:id, 1)
```
paulcsmith: If you send the model with the id on it already that could clean up the assertions and make them a bit more robust

```
model = %{model | id: 1}
send self, {:create, model}
model
```

paulcsmith:
Thanks for this. I pulled it and looks like all tests are passing. LMK what you think of those comments, and then we can merge this in! Thank you very much @jesenko  :D

test/ex_machina/ecto_test.exs
jesenko: I agree, will update PR asap...

test/ex_machina/ecto_test.exs
jesenko: Combined with change you described bellow (i.e. changing the order of adding `id` and sending model), this would work...

The rationale I originally had for sending model _before_ adding id was to be able to test what is passed to `Ecto.Repo`. In _changeset_ PR from which this was extracted, `Ecto.Changeset` was passed to `Repo.Insert`, while `Ecto.Model` was returned as `created_record`.

Currently, no preprocessing is done on `record` passed to `Repo.Insert`, so not much is to test here, so I am :+1: on this... It will not be a hassle to switch order again if something worthy of testing comes up in future.

paulcsmith: Yeah that is totally understandable! Good points. I agree, if we need to test that scenario we can always change it later. Good work on this!

jesenko:
I have updated the PR accordingly, should be OK to merge now...

Thank you for great library!

jesenko:
@paulcsmith @jsteiner In future, it might be better to test agains some actual `Ecto.Repo` implementation, so possible issues in struct passed from ExMachina to Ecto can be detected. E.g., if you now specify `:has_many` associations in `factory`, tests will pass just fine, however if you actually do that, `Ecto.Repo` will fail as it does not accept `has_many` assocs in `insert!`.

After inspecting how testing is done in `Ecto` library, probably the best approach would be to have something like [this](https://github.com/elixir-lang/ecto/blob/master/test/support/test_repo.exs). `/test/` folder is however not exported in `ecto` package, so it is not possible to just require it from `/deps/` folder. Easiest way would probably be just to copy it into this repo and then keeping it in sync with ecto version...

Another option would be to introduce test dependency on `sqlite` adapter and test against that, however that might be overkill...

What do you think?

paulcsmith:
@jesenko I was thinking the same thing. It would be good to have some integration tests that actually run through Ecto to make sure things are working as expected. Thanks for the link to Ecto's TestRepo. We might try something like that in the future.

We'll take a final look at this and merge it in soon. Thanks for working on this! :D

paulcsmith:
Rebased and merged into `0.3.0` branch by https://github.com/thoughtbot/ex_machina/commit/579913040f7cd6cce810a12f39635416158e2adb

PR #21, opened 2015-09-25
Allow chaining with built records

paulcsmith:
This will allow things like:

```
build(:user, name: "john") |> on_free_trial |> create |> add_comment(title: "toto")
```

So that you can easily modify the built record, then call create and do `has_many/has_one` relationships with the saved record. I will be adding the functionality to not save associations on build in the next PR, which should make this a bit more robust


BlakeWilliams:
LGTM

paulcsmith:
Thanks!

PR #25, opened 2015-10-02
Improve error when no repo is given

paulcsmith:
Right now the error message is not that helpful. This makes it a bit
more clear what is wrong, and how to resolve the issue.


jsteiner:
:+1:

PR #26, opened 2015-10-02
Cleaning up warnings

### LINKED ISSUE #22: warning: this clause cannot match because a previous clause always matches
There is a function that throws this warning. I believe it is the save_record function because we overwrite it in the custom module, when calling `use ExMachina.Ecto` we should figure out where that is, and how to get rid of the warning. Maybe with defoverridable?

paulcsmith:
* Removes warning when defining custom save_record/1 function
* Removes unnecessary `module` param in `ExMachina.Ecto.save_record`
* Corrects docs. `save_function/1` to `save_function/2`

Closes #22

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ end
 ## Using without Ecto
 
 You can use ExMachina without Ecto, by using just the `build` function, or by
-defining `save_function/1` in your module.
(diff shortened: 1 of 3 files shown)


jsteiner:
:+1:

PR #27, opened 2015-10-02
Add note about importing with Phoenix

paulcsmith:
Even if the Phoenix team decides not to import `build` by default it might make sense to include this since people that already have a generated Phoenix application would have the `Eco.Model` import.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ import Ecto.Model, except: [build: 2]
 
 ```elixir
 # test/factories.ex
-defmodule MyApp.Factories do
+defmodule MyApp.Factory do
   # MyApp.Repo is an Ecto Repo.
   # It will automatically be used when calling `create`
   use ExMachina.Ecto, repo: MyApp.Repo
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -25,11 +25,28 @@ defp app_list(_),  do: app_list
...
 
+## Using with Phoenix and Ecto
+
+There is nothing special you need to do with Phoenix unless you decide to
+`import` your factory module.
+
+By default Phoenix `import`s `Ecto.Model` in the generated `ConnCase` and
+`ModelCase`  modules (found in `test/support/conn_case.ex` and
+`test/support/model_case.ex`). To import your factory we recommend excluding
+`build/2` or aliasing your factory instead.
```
jsteiner: Maybe add a note: "We recognize that this is not ideal, and are currently working on a solution."

jsteiner:
One comment. Otherwise :+1:

README.md
paulcsmith: I'm not sure that adds value. To me it makes the library look incomplete and shady, without giving me any valuable information IMO :S

I'm leaning toward updating the README when we have a solution and just leaving this as-is in the meantime. What do you think?

PR #28, opened 2015-10-02
Move functions from use to before compile

jsteiner:
This should make it easier to override each function within your own
test suite.


paulcsmith:
If all tests pass I think this is good to merge in :+1:

PR #30, opened 2015-10-02
Don't save associations in build

jsteiner:
This PR makes it so records, including associations, are only saved in `save_record`. There is no reason to save the records on the build step, so this will save unnecessary writes, while making build more intuitive.


paulcsmith:
I think this is good to merge in. I realized we also need to update the `@version` in mix.exs at some point.

PR #32, opened 2015-10-07
Add a useful error message when you try to save a non-Ecto struct with ExMachina.Ecto

jsteiner:


test/ex_machina/ecto_test.exs
```diff
@@ -127,6 +127,16 @@ defmodule ExMachina.EctoTest do
     assert TestRepo.one(User)
   end
 
+  test "save_record/1 raises for associated records that are not Ecto structs" do
+    author = %{}
+
+    message = "expected association: `%{}` at key: `:author` to be an Ecto struct."
```
paulcsmith: What do you think of `Expected :author to be an Ecto struct, but got %{}`?

That follows the some of the other errors I've seen in Ecto: https://github.com/elixir-lang/ecto/search?utf8=✓&q=got&type=Code

paulcsmith:
Just had one comment/suggestion. Looks good though!

test/ex_machina/ecto_test.exs
jsteiner: Ahh, much better!

PR #34, opened 2015-10-09
Ignore associations that are not set

jsteiner:
This fixes an error where a factory has an association, but the record
is not set in the factory definition or passed in when creating the
factory. Since associations are structs of type
`Ecto.Association.NotLoaded`, we can ignore these and return the record
as is.

### DIFF
--- a/test/ex_machina/ecto_test.exs
+++ b/test/ex_machina/ecto_test.exs
@@ -127,6 +127,13 @@ defmodule ExMachina.EctoTest do
     assert TestRepo.one(User)
   end
 
+  test "save_record/1 ignores associations that are not set" do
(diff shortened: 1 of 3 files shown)


test/ex_machina/ecto_test.exs
```diff
@@ -127,6 +127,13 @@ defmodule ExMachina.EctoTest do
     assert TestRepo.one(User)
   end
 
+  test "save_record/1 ignores associations that are not set" do
+    article = Factory.save_record(%Article{title: "Ecto is Awesome"})
```
paulcsmith: What do you think about explicitly passing `Ecto.Association.NotLoaded` as the author? That way it's super clear what it's testing?

paulcsmith:
Just one small comment

PR #36, opened 2015-10-09
Fix association bug

paulcsmith:
Before this was only setting the association's id when saving a built
record, but that meant that the association still had the built record,
not the newly saved one. This makes sure that `put_assoc` set both the
id, and the associated record


jsteiner:
:+1:

PR #39, opened 2015-10-23
Return to using functions for factory definitions

jsteiner:
Functions have less complexity and less magic than normal functions.

The functional approach also provides access to the `attrs` hash, which
allows you to utilize it in your factory definition:

    def factory :user, attrs do
      modified_attrs = attrs |> modify_attrs
      %{
        name: "Foobar",
        accounts: assoc(modified_attrs, :accounts),
      }
    end

### DIFF
--- a/lib/ex_machina.ex
+++ b/lib/ex_machina.ex
@@ -73,13 +73,19 @@ defmodule ExMachina do
     end
   end
 
-  defmacro factory(factory_name, do: block) do
-    quote do
-      def factory(unquote(factory_name), var!(attrs)) do
-        !var!(attrs) # Removes unused variable warning if attrs wasn't used
-        unquote(block)
+  defmacro factory(factory_name, do: _block) do
+    raise """
+    The factory and assoc macros have been removed. Please use regular
+    functions instead.
+
+      def factory(#{factory_name}, attrs) do
+        %{
+          ...
+          some_assoc: assoc(attrs, :some_assoc)
+          ...
+        }
       end
-    end
+    """
   end
 
   @doc """
@@ -87,7 +93,7 @@ defmodule ExMachina do
 
   ## Examples
 
-      factory :user do
+      def factory(:user, _attrs) do
         %{
           # Will generate "me-0@example.com" then "me-1@example.com", etc.
           email: sequence(:email, &"me-\#{&1}@foo.com")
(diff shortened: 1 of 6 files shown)


lib/ex_machina.ex
```diff
@@ -150,7 +141,7 @@ defmodule ExMachina do
 
   ## Example
 
-      factory :user do
+      def factory :user, _attrs do
```
paulcsmith: I think we should add parens around the parameters. I think that's a common convention

lib/ex_machina/ecto.ex
```diff
@@ -5,14 +5,16 @@ defmodule ExMachina.Ecto do
...
 
-        import ExMachina.Ecto, only: [assoc: 1, assoc: 2]
-
         @repo unquote(repo)
 
         def fields_for(factory_name, attrs \\ %{}) do
           ExMachina.Ecto.fields_for(__MODULE__, factory_name, attrs)
         end
 
+        defp assoc(attrs, factory_name, opts \\ []) do
```
paulcsmith: Since this is defp

lib/ex_machina.ex
```diff
@@ -73,21 +73,12 @@ defmodule ExMachina do
     end
   end
 
-  defmacro factory(factory_name, do: block) do
```
paulcsmith: I'm wondering if rather than removing this we should add raise a helpful error on the `factory` and `assoc` macros. Just for this version so that it makes an upgrade easy. I want the upgrade process to be as smooth as possible. A lot of people upgrade without looking at release notes so I think something like this would help:

```elixir
# I haven't tested this so it may not work exactly as-is
defmacro factory(factory_name, _attrs) do
  raise """
  The factory macro has been removed. Please use a regular function instead.

    def factory(#{factory_name}, attrs) do
        ...
    end
  """
end
```

paulcsmith:
Left just a couple comments :+1:

lib/ex_machina.ex
```diff
@@ -73,13 +73,19 @@ defmodule ExMachina do
...
-        unquote(block)
+  defmacro factory(factory_name, do: _block) do
+    raise """
+    The factory and assoc macros have been removed. Please use regular
+    functions instead.
+
+      def factory(#{factory_name}, attrs) do
+        %{
+          ...
+          some_assoc: assoc(attrs, :some_assoc)
```
paulcsmith: Yeah this probably works for both :+1:

paulcsmith:
LGTM

PR #42, opened 2015-10-23
Add CircleCI config

jsteiner:


paulcsmith:
I wonder if we can add: `mix compile --warnings-as-errors` in there somewhere? I think it would be bad to release a version that produces warnings

paulcsmith:
I think that didn't work because I ran it locally and there is a failure :S

= Compilation error on file lib/ex_machina/ecto.ex ==
** (CompileError) lib/ex_machina/ecto.ex:109: Ecto.Association.BelongsTo.__struct__/0 is undefined, cannot expand struct Ecto.Association.BelongsTo
    (elixir) src/elixir_map.erl:58: :elixir_map.translate_struct/4
    (stdlib) lists.erl:1353: :lists.mapfoldl/3
    (stdlib) lists.erl:1354: :lists.mapfoldl/3
    (elixir) src/elixir_for.erl:101: :elixir_for.translate_gen/5
    (elixir) src/elixir_for.erl:81: :elixir_for.translate_gen/4

paulcsmith:
Also, we should definitely figure out how to add it because it caught some problems with Ecto has_one/many commit. Our code is using the `%Ecto.Association.BelongsTo{}` form, instead of `%{__struct__: Ecto.Association.BelongsTo}` so it fails to compile without Ecto

paulcsmith:
I put this in Slack, but will add it here as well for history/searchability

> I think I know why it passed, it was compiled in the test environment (which includes the Ecto dep)

> So we may need to do `MIX_ENV=prod mix compile --warnings-as-errors`

PR #43, opened 2015-10-23
Always compile Ecto

jsteiner:
ExMachina may be used in other environments.


paulcsmith:
I don't think this is the solution. We shouldn't require Ecto as a dependency, instead I think we want to change all the `%Ecto...{}` structs to `%{__struct__: Ecto...}`

jsteiner:
Ah, yeah, that makes more sense. I see that now in the other PR. Doing a not so great job at multitasking...

paulcsmith:
Haha it's all good. Multitasking is hard :)

jsteiner:
Merged.

PR #45, opened 2015-10-29
Fix intermittent failure

### LINKED ISSUE #44: Fix intermittently failing sequence test
I believe it's because the test is running async. If we then that off we should be fine. I'll tackle that tomorrow

paulcsmith:
Some of the other tests are running asynchronously and were causing
issues if they both used the same sequence name (in this case `email`).
Running the tests synchronously should fix the issue.

The tests are still plenty fast. Once Ecto supports concurrent tests, we may want to rethink the Sequence module so that it keeps sequences *per process* so others don't run into this issue with their tests. Fixes #44


jsteiner:
:+1:

PR #48, opened 2015-10-30
Tests for saving embedded Ecto models

iancanderson:
This covers behavior added in 080498ee4c3a5a70b69f478c7877d2979e879e01

Also updated mix.exs so that a migration could be generated during development without prefixing with MIX_ENV=test.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -36,7 +36,7 @@ defmodule ExMachina.Mixfile do
     [
       {:ex_doc, "~> 0.9", only: :dev},
       {:earmark, ">= 0.0.0", only: :dev},
-      {:ecto, "~> 1.0", only: :test},
+      {:ecto, "~> 1.0", only: [:dev, :test]},
       {:sqlite_ecto, "~> 1.0.0", only: :test}
     ]
   end
--- a/test/ex_machina/ecto_embeds_test.exs
+++ b/test/ex_machina/ecto_embeds_test.exs
@@ -1,0 +1,55 @@
+defmodule ExMachina.EctoEmbedsTest do
+  use ExMachina.EctoCase
+  alias ExMachina.TestRepo
+
+  defmodule User do
+    use Ecto.Model
+    schema "users" do
+      field :name, :string
+      field :admin, :boolean
+      embeds_one :settings, ExMachina.EctoEmbedsTest.Settings
+    end
+  end
+
+  defmodule Settings do
+    use Ecto.Model
+    embedded_schema do
+      field :email_signature
+      field :send_emails, :boolean
+    end
+  end
+
+  defmodule Factory do
+    use ExMachina.Ecto, repo: TestRepo
+
+    def factory(:settings, _attrs) do
+      %Settings{
+        email_signature: "Mr. John Doe",
+        send_emails: true
+      }
+    end
+
+    def factory(:user, _attrs) do
+      %User{
+        name: "John Doe",
+        admin: false,
+        settings: build(:settings)
+      }
+    end
+  end
+
+  test "create/1 saves `embeds_one` record defined in the factory" do
+    Factory.create(:user)
+
+    user = TestRepo.one(User)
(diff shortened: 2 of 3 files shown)


mix.exs
```diff
@@ -36,7 +36,7 @@ defmodule ExMachina.Mixfile do
     [
       {:ex_doc, "~> 0.9", only: :dev},
       {:earmark, ">= 0.0.0", only: :dev},
-      {:ecto, "~> 1.0", only: :test},
+      {:ecto, "~> 1.0", only: [:dev, :test]},
```
paulcsmith: NIIICE :+1:

test/ex_machina/ecto_embeds_test.exs
```diff
@@ -0,0 +1,55 @@
...
+        settings: build(:settings)
+      }
+    end
+  end
+
+  test "create/1 saves `embeds_one` record defined in the factory" do
+    Factory.create(:user)
+
+    user =  TestRepo.one(User)
+    assert %{settings: %{email_signature: "Mr. John Doe", send_emails: true}} = user
```
paulcsmith: I think we should still keep that test that makes sure the state is set to `loaded` with `user.settings.__meta__.state == :loaded`

test/ex_machina/ecto_embeds_test.exs
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  test "create/1 saves `embeds_one` record when overridden" do
```
paulcsmith: `create/2` in the test description

test/ex_machina/ecto_embeds_test.exs
iancanderson: But the user is being loaded from `TestRepo`, so why would it not be `:loaded`? Isn't that straight up testing `Ecto` at that point?

test/ex_machina/ecto_embeds_test.exs
iancanderson: whoops, nice catch

test/ex_machina/ecto_embeds_test.exs
```diff
@@ -0,0 +1,55 @@
...
+    settings = %Settings{email_signature: "Mrs. Jane Doe", send_emails: false}
+    Factory.create(:user, settings: settings)
+
+    user =  TestRepo.one(User)
+    assert %{settings: settings} = user
```
paulcsmith: I *think* this should be `%{settings: ^settings}` with a pinned `^settings` variable. Otherwise I think it is assigning to the variable `settings`

A better solution might be to do

```
assert user.settings.email_signature == settings.email_signature
```

and then not override `send_emails` on line 49. What do you think?

test/ex_machina/ecto_embeds_test.exs
```diff
@@ -0,0 +1,55 @@
...
+    user = TestRepo.one(User)
+    assert %{settings: %{email_signature: "Mr. John Doe", send_emails: true}} = user
+  end
+
+  test "create/2 saves `embeds_one` record when overridden" do
+    settings = %Settings{email_signature: "Mrs. Jane Doe", send_emails: false}
+    Factory.create(:user, settings: settings)
+
+    user = TestRepo.one(User)
+    assert %{settings: settings} = user
```
paulcsmith: Reposting because it was buried after a push

I think this should be %{settings: ^settings} with a pinned ^settings variable. Otherwise I think it is assigning to the variable settings

A better solution might be to do

```
assert user.settings.email_signature == settings.email_signature
```

and then not override send_emails on line 49. What do you think?

test/ex_machina/ecto_embeds_test.exs
paulcsmith: Good point! Carry on haha

paulcsmith:
This is good to go. Just one comment about the second test

paulcsmith:
Thanks for tackling this. I think we're good to publish after this is in :D

iancanderson:
@paulcsmith happy to help! NRR

PR #49, opened 2015-10-30
Add associations documentation

paulcsmith:


README.md
```diff
@@ -44,21 +44,24 @@ defmodule MyApp.Factory do
   def factory(:user, _attrs) do
     %User{
       name: "Jane Smith",
-      email: sequence(:email, &"email-#{&1}@example.com"),
```
jsteiner: :(

jsteiner: I typically prefer leaving these on because it makes it easier to reorder lines without having to add and remove commas. It also makes diffs less noisy.

jsteiner:
One comment, re: trailing commas. Otherwise :+1:

README.md
paulcsmith: For some reason I thought these weren't allowed in Elixir. I'll change it back :)

PR #53, opened 2015-11-08
README: suggest location for factory.ex

henrik:
This was not obvious to me as a Phoenix newbie and a new user of this lib.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ In `mix.exs`, add the ExMachina dependency:
 
 ```elixir
 def deps do
-  [{:ex_machina, "~> 0.4"}]
+  [{:ex_machina, "~> 0.4", only: "test"}]
 end
 ```
 
@@ -27,6 +27,19 @@ defp app_list(_),  do: app_list
 defp app_list,  do: [:logger]
 ```
 
+Make sure that the `test/support` directory (where you'll put your factory definitions) will be compiled. If you use Phoenix, this should already be done for you.
+
+```elixir
+def project do
+  […
+   elixirc_paths: elixirc_paths(Mix.env),
+  …]
+end
+
+defp elixirc_paths(:test), do: ["lib", "test/support"]
+defp elixirc_paths(_), do: ["lib"]
+```
+
 ## Cheatsheet
 
 Check out [the docs](http://hexdocs.pm/ex_machina/ExMachina.html) for more details.
@@ -34,6 +47,8 @@ Check out [the docs](http://hexdocs.pm/ex_machina/ExMachina.html) for more detai
 Define factories:
 
 ```elixir
+# You might put this in `test/support/factory.ex`.
+
 defmodule MyApp.Factory do
   # with Ecto
   use ExMachina.Ecto, repo: MyApp.Repo


README.md
```diff
@@ -34,6 +34,9 @@ Check out [the docs](http://hexdocs.pm/ex_machina/ExMachina.html) for more detai
 Define factories:
 
 ```elixir
+# You could put this in `test/support/factory.ex` if you use it for tests only,
```
paulcsmith: One problem with this is that I believe that `test/support` is not in the `elixirc_paths` by default. In phoenix it is, but a regular mix project doesn't include that as far as I know. Any ideas for that? Maybe something like

```
# You could put this in `test/factory.ex` or in `lib/my_app/factory.ex` if you want to use it for development seed data as well.
# If using Phoenix, you may want to put it in `test/support/factory.ex`

**Comment by paulcsmith** [comment]
Thanks for much for this contribution! I left just one small comment

File: README.md (hunk 1 above)
Comment (henrik) [inline README.md]: Oh yeah, I used this in a Phoenix project – didn't think of the greater world. Sounds like a good change.

Also, I'm not sure about the `lib` recommendation – I just got this trying to deploy my app:

>        ** (CompileError) lib/toolbox/factory.ex:3: module ExMachina.Ecto is not loaded and could not be found

(that's the line `use ExMachina.Ecto, repo: Toolbox.Repo`)

This is likely because my mix.exs says

``` elixir
 {:ex_machina, "~> 0.4", only: [:dev, :test]},
```

So in prod, this file which *is* in `elixirc_paths` does not have its dependency.

`test/support` is cleverly only in the path in tests. I'll probably remove the lib recommendation (I don't use it in dev anymore myself) – it could be solved but it might not be a very common case.

henrik: Hm, seems the default `elixirc_paths` is just `lib`. I tried adding an .ex file to `test` in a new project (non-Phoenix) and it wasn't available to use from tests.

So I guess if you've used this outside Phoenix, you made sure to add whatever directory it was to `elixirc_paths`? If so we could doc that.

paulcsmith: @henrik What if you do `test/factory.exs`. I think it might beed to be a scripted elixir file. Not sure though

henrik: Tried a scripted file as well, didn't seem to work. I don't think you can use a module defined in a script from another script – I think the stuff you use elsewhere needs to be compiled, and scripts are just things that can use compiled things but can't be used themselves.

Experiments seem to bear that out but I could be wrong.

I force pushed an update, suggesting adding the `test/support` dir to the path if it's not there.

README.md
```diff
@@ -11,7 +11,9 @@ In `mix.exs`, add the ExMachina dependency:
 
 ```elixir
 def deps do
-  [{:ex_machina, "~> 0.4"}]
+  [
+    {:ex_machina, "~> 0.4", only: "test"}
+  ]
```
paulcsmith: I like this change, but I think that it should remain on one line for the README. To keep things a bit more concise

README.md
```diff
@@ -19,7 +21,9 @@ Add `:ex_machina` to your application list:
 
 ```elixir
 def application do
-  [applications: app_list(Mix.env)]
+  [
+    applications: app_list(Mix.env)
```
paulcsmith: One line here as well

README.md
```diff
@@ -11,7 +11,9 @@ In `mix.exs`, add the ExMachina dependency:
 
 ```elixir
 def deps do
-  [{:ex_machina, "~> 0.4"}]
+  [
+    {:ex_machina, "~> 0.4", only: "test"}
```
henrik: @paulcsmith I really wanted to add a trailing comma here and elsewhere – it's a strongly preferred style for me for many reasons – but I suspect you don't want that here? I don't think it's idiomatic in Elixir, sadly.

I checked thoughtbot's "guides" repo, and it *is* in your Ruby guide, but you don't have an Elixir one…

README.md
```diff
@@ -27,6 +31,20 @@ defp app_list(_),  do: app_list
 defp app_list,  do: [:logger]
 ```
 
+Make sure that the `test/support` directory (where you'll put your factory definitions) will be compiled:
```
paulcsmith: Maybe a not that this is already done in Phoenix by default?

README.md
paulcsmith: @henrik Good catch. We typically do that in the code itself, but in the README, I think it just needs to show you how to add it to your project in a concise way.

henrik:
Force pushed once more to combine the "apps" and "elixirc_path" bit into a single example.

README.md
henrik: I like commas in the README so you can add it anywhere in your project without modification, but I think it might be a WTF factor to those who haven't seen the light ;)

README.md
```diff
@@ -11,20 +11,28 @@ In `mix.exs`, add the ExMachina dependency:
...
 
-Add `:ex_machina` to your application list:
+Add `:ex_machina` to your application list and also make sure that the `test/support` directory (where you'll put your factory definitions) will be compiled:
 
 ```elixir
 def application do
-  [applications: app_list(Mix.env)]
+  [
+    applications: app_list(Mix.env),
+    elixirc_paths: elixirc_paths(Mix.env)
```
paulcsmith: I believe that the `elixirc_paths` are set in the `project`. Does this also work?

README.md
henrik: My thinking was that I always copy-paste these, and it's more fiddly if it's condense. But if you still prefer that style, let me know and I'll change it.

paulcsmith: Yeah I think condensed is preferred for now, but we might change it in the future. I'll have to think about it some more.

README.md
henrik: No, you're right, my mistake. Good catch, thank you.

henrik: Fixed, force-pushed. Also re-condensed the deps line.

README.md
henrik: Oh, missed this bit. Yeah, I considered that. Will add.

README.md
henrik: Done.

henrik:
Think I've fixed all the comments. Please let me know what you think.

paulcsmith:
Thanks Henrik, still thinking about whether this is the best thing to recommend or not. I'd like to look at some other tools to see what is most common. I'm beginning to think the default recommendation should be as simple as possible. Add the ex_machina dep in all environments and then add it to `lib/factory.ex`. Then if people want they can add to `test/support`, but the default is really simple.

This might be good because it makes it simple to get started, makes it easy to use to generate seed data, and is less documentation for us to maintain in the future. What do you think?

henrik:
That sounds good to me. lib/factory.ex or lib/myproject/factory.ex? Former is simpler and doesn't include a fake value to be substituted. I guess Elixir code doesn't hold as strongly to the "file path matches name" convention as Rails code has to.

henrik:
No need to go through me/this PR if you just add a suggested path btw - have at it :)

eidge:
+1 on this PR, I was having exactly the same problem.

Regarding the file structure, is it your advise that all the factories live inside the same file, or one file per model?

paulcsmith:
@jsteiner what do you think about making the default suggestion to add the dependency in all environments and then add the factory module to lib. That way it doesn't require that much change to get started and you can use factories for seed data as well. 

Honestly I think this is my preferred way to do it now anyway since it's easier and it's nice to be able to use factories for seeds. Thoughts?

paulcsmith:
@eidge thanks for bringing this up again. We'll figure out the preferred way and add it to the Readme. 

I think it's much better to keep all factories in one module. I'll write a blog post soon on how to structure the modules and corresponding functions. For now, just putting them in one module works pretty well :)

jsteiner:
@paulcsmith I've found factories useful for seed data as well, so that seems fair.

paulcsmith:
Thanks for your help on this @henrik. We took a slightly different approach, but I'm glad you brought this to our attention! Closed by #67

henrik:
@paulcsmith No worries, glad things ended up clearer :)

PR #55, opened 2015-11-09
Update instructions to start ex_machina in the test env

scrogson:
Users should start `ex_machina` in `test/test_helper.exs` rather than modifying the applications list in `mix.exs`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,16 +15,11 @@ def deps do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


paulcsmith:
Oh I like this a lot more. Thanks, @scrogson. Any other feedback you have on ExMachina?

PR #57, opened 2015-11-09
Remove factory/2 and assoc/3

jsteiner:
The `attr` param in `factory/2` could be confusing. Now that we automatically persist all `belongs_to` associations when `create`ing records, `attrs` is unnecessary as we can simply `build` a record and let `ExMachina` take care of saving the record as it would by default.

Now, we call `factory/1` which is easier to understand, and you no longer have to pass `attrs` when you don't need it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ ExUnit.start()
 Application.ensure_all_started(:ex_machina)
 ```
 
-## Cheatsheet
+## Overview
 
 [Check out the docs](http://hexdocs.pm/ex_machina/ExMachina.html) for more details.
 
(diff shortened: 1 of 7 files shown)


lib/ex_machina/ecto.ex
```diff
@@ -15,8 +15,18 @@ defmodule ExMachina.Ecto do
           ExMachina.Ecto.save_record(__MODULE__, @repo, record)
         end
 
-        defp assoc(attrs, factory_name, opts \\ []) do
-          ExMachina.Ecto.assoc(__MODULE__, attrs, factory_name, opts)
+        defp assoc(_, factory_name, _) do
+          raise """
+          assoc/3 has been removed. Please use build instead.
```
paulcsmith: Would it be helpful to explain that build associations will be saved when you use `create`?

README.md
```diff
@@ -36,27 +36,27 @@ defmodule MyApp.Factory do
...
     }
   end
 
-  def factory(:article, attrs) do
+  def factory(:article) do
     %Article{
       title: "Use ExMachina!",
       # has_many/has_one associations are inserted when you call `create`
       comments: [build(:comment)],
       # only available in ExMachina.Ecto
```
paulcsmith: Maybe this can be changed to `# ExMachina.Ecto will insert the record if `create/2` is called?

paulcsmith:
I really like this change. I think it would be good to add a section to the README on associations as well. Maybe in a separate commit. It could detail how associations work, when they're inserted and advise people to use `build` and *not* `create` inside of factory definitions. Using `create` could lead to performance issues and bugs, but many people might not realize that

README.md
```diff
@@ -36,27 +36,27 @@ defmodule MyApp.Factory do
...
-      author: assoc(attrs, :author, factory: :user),
```
paulcsmith: SOOO much nicer

jsteiner: Yep. And now, you can even override the default object with whatever attributes you want.

paulcsmith: Oh yeah. I hadn't thought of that! That's a nice plus

paulcsmith:
Looks awesome! Let's get this in :D

jsteiner:
Merged into 0.5.0.

PR #58, opened 2015-11-10
use always correct association_id key name (as defined on schema)

mindreframer:
Problem: 
- current implementation works based on convention, that association names always follow the pattern "#{association_name}_id". This is not always the case

Solution: 
- use the information on schema for that association to get the owner key

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -172,9 +172,12 @@ defmodule ExMachina.Ecto do
     end
   end
 
-  defp put_assoc(record, association_name, association) do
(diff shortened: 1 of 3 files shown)


paulcsmith:
@mindreframer Good catch and thanks for the PR! Could you add some tests for this? If you need some help, let me know

mindreframer:
@paulcsmith here's a small unit test for that case. Hope it's all right now. Cheers and thanks for ex_machina!

paulcsmith:
Merged! Thanks for your help @mindreframer :)

mindreframer:
@paulcsmith I have to thank! Great library and quite helpful!

PR #59, opened 2015-11-11
Add docs about what `attrs` are used for

paulcsmith:
Someone mentioned they didn't realize that what the second attrs were until they tried it. I think it would be helpful to add something to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Application.ensure_all_started(:ex_machina)
 
 ## Cheatsheet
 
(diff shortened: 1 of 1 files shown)


jsteiner:
:+1:

BlakeWilliams:
LGTM, thoughts on putting them at the bottom of each section instead of at the top?

paulcsmith:
@BlakeWilliams That's a good thought. I don't have a strong preference either way. I might try it out later and see how it looks

PR #60, opened 2015-11-13
Fix test data to match with the test description

scrogson:
This test says its supposed to merge a map, but it was a keyword list.


paulcsmith:
Oops :) Thanks @scrogson

PR #61, opened 2015-11-13
Drop `nil` values in `fields_for/2`

scrogson:
When using `fields_for/2` while testing controllers or changesets, you end up having to manually remove unwanted keys  to satisfy your tests. This ensures that keys with `nil` values are removed.

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -56,6 +56,7 @@ defmodule ExMachina.Ecto do
   def fields_for(module, factory_name, attrs \\ %{}) do
     module.build(factory_name, attrs)
     |> drop_ecto_fields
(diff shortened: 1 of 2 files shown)


jsteiner:
Might there be situations where it would be useful to set a value to be nil? What do you think about dropping `nil` values _before_ merging in `attrs`?

So, something like this:

```elixir
assert Factory.fields_for(:user, name: nil) == %{
  # no id: nil
 name: nil,
 admin: false
}
```

paulcsmith:
I think it makes sense to keep nil values. If you want to strip nil values, maybe you could add a new function to your factory that calls `fields_for/2` and then strips them out. Feel free to post here or open a new issue if you have more ideas or another suggestion though

PR #62, opened 2015-11-13
Improve mix description

jsteiner:
ExMachina currently doesn't show up on hex.pm if you search for "factory" or
"factories". I'd like to at least appear on that list. There are some
other libraries that do show up and mention that they are clones of
FactoryGirl. That seems to be something people are interested in, so it
may be worth saying that this is from the same people who created
FactoryGirl.


paulcsmith:
I think it's worth the experiment to see how it affects traffic!

Issue #63, opened 2015-11-16
Raise a helpful error letting people know to define factory/1 instead of factory/2


paulcsmith:
We added an error here: https://github.com/thoughtbot/ex_machina/blob/0.5.0/lib/ex_machina.ex#L212

but that will never be called since internally we are calling factory/1. Instead that error message should be added to https://github.com/thoughtbot/ex_machina/blob/0.5.0/lib/ex_machina.ex#L16 `UndefinedFactory`. Something like:

```elixir
"""
No factory defined for #{inspect factory_name}. This may be because you have defined a factory with attrs like this: def factory(:my_factory, attrs)

This has been changed in 0.5.0 to just be: def factory(:my_factory). The assoc/2 function has also been removed. Please define your factory function like this:

def factory(#{inspect factory_name}) do
  ...
end
"""
```

rranelli:
I think this issue can be closed, right?

paulcsmith:
@rranelli Yes thank you for pointing that out. Slipped right by :)

PR #64, opened 2015-11-16
Actually display error message

jsteiner:
Previously, the error message defined in `factory/2` would not be used
because internally we were only calling `factory/1`. This improves the
error message and moves it to the appropriate spot.

Addresses: https://github.com/thoughtbot/ex_machina/issues/63

### DIFF
--- a/lib/ex_machina.ex
+++ b/lib/ex_machina.ex
@@ -13,7 +13,27 @@ defmodule ExMachina do
     defexception [:message]
 
     def exception(factory_name) do
-      message = "No factory defined for #{inspect factory_name}"
+      message =
+        """
+        No factory defined for #{inspect factory_name}. This may be because you
+        defined a factory with two parameters like this:
+
+            def factory(:my_factory, attrs)
+
+        As of ExMachina 0.5.0, we no longer call factory/2. Please define your
+        factory function like this:
+
+            def factory(#{inspect factory_name}) do
+              ...
+            end
+
+        The assoc/3 function has also been removed. belongs_to relationships
+        can now be used with build:
+
+            def factory(#{inspect factory_name}) do
+              parent: build(:parent)
+            end
+        """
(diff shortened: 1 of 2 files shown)


lib/ex_machina.ex
```diff
@@ -13,7 +13,25 @@ defmodule ExMachina do
...
+
+        This has been changed in 0.5.0 to be: def factory(:my_factory). The
+        assoc/2 function has also been removed. Please define your factory
+        function like this:
+
+        def factory(#{inspect factory_name}) do
+          ...
+        end
+
+        belongs_to relationships can now be defined with build:
```
paulcsmith: Maybe "defined" -> "used"?

lib/ex_machina.ex
```diff
@@ -13,7 +13,25 @@ defmodule ExMachina do
     defexception [:message]
 
     def exception(factory_name) do
-      message = "No factory defined for #{inspect factory_name}"
+      message =
+        """
+        No factory defined for #{inspect factory_name}. This may be because you
+        defined a factory with attrs like this: def factory(:my_factory, attrs)
```
paulcsmith: Maybe this could be clarified a bit

"defined a factory with 2 parameters like this:"

lib/ex_machina.ex
```diff
@@ -13,7 +13,25 @@ defmodule ExMachina do
...
+
+        This has been changed in 0.5.0 to be: def factory(:my_factory). The
+        assoc/2 function has also been removed. Please define your factory
```
paulcsmith: Maybe this line ("the assoc/3 function has als been removed") to line 29?

"The assoc/3 function has also been removed. belongs_to relationships can now be defined with build:"

paulcsmith:
:+1: I think we can test this on our apps and if all passes we should release. This is a pretty awesome change that will make it a lot easier to get started I think

PR #67, opened 2015-12-07
README - Where to put factories

paulcsmith:
As discussed in #53

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,15 @@ def deps do
 end
 ```
 
-Start `:ex_machina` for the test environment in `test/test_helper.exs`:
+And be sure to start the ExMachina application. For most projects (such as
+Phoenix apps) this will mean adding `:ex_machina` to the list of applications in
+`mix.exs`.
 
 ```elixir
-ExUnit.start()
-Application.ensure_all_started(:ex_machina)
+def application do
+  [mod: {MyApp, []},
+   applications: [:ex_machina, :other_apps...]]
+end
 ```
 
 ## Overview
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,6 +88,16 @@ create_list(3, :comment, attrs)
 fields_for(:comment, attrs)
 ```
 
+## Where to put your factories
+
+We recommend starting by creating one factory module (such as MyApp.Factory) in
+`lib/my_app/factory.ex` and putting all factory definitions in that module.
+
+Later on you can easily create different factories if you need to hook them up
+to different repos, if your factory module is getting to big, or if you have
```
BlakeWilliams: Thoughts on ending the first sentence at `to different repos.` Then maybe something along the lines of: `If you prefer multiple factories for any reason you can create another factory module in the same directory`.

I think it would help if `create another factory module` was more clearly defined or an example was given to avoid confusion.

BlakeWilliams:
One comment, otherwise LGTM

README.md
paulcsmith: Yeah you're right. That's a confusing sentence. I'll rework it :+1:

README.md
```diff
@@ -84,6 +88,16 @@ create_list(3, :comment, attrs)
 fields_for(:comment, attrs)
 ```
 
+## Where to put your factories
+
+We recommend starting by creating one factory module (such as MyApp.Factory) in
+`lib/my_app/factory.ex` and putting all factory definitions in that module.
+
+Later on you can easily create different factories by creating a new module in
+the same directory. This can be helpful if you need to create factories that are
```
paulcsmith: @BlakeWilliams What do you think of this?

BlakeWilliams: I like it!

eidge:
Nice, thanks. That's really useful :)

Issue #70, opened 2015-12-15
Ecto 1.1 Changes a bunch of things


geofflane:
The breaking change is is `Ecto.Association.loaded?` is no longer a function, but they've added `Ecto.assoc_loaded?`  as a public function.

There are some deprecation warnings too from changes in the handling of assoc, embed, etc inside of changesets is different now and causes warnings in ex_machina right now.

```
.warning: changing assocs with change/2 or put_change/3 is deprecated, please use put_assoc/3 instead
    (stdlib) lists.erl:1262: :lists.foldl/3
    (ecto) lib/ecto/changeset.ex:247: Ecto.Changeset.change/2
    (ex_machina) lib/ex_machina/ecto.ex:128: ExMachina.Ecto.save_record/3
    test/ex_machina/ecto_has_many_test.exs:77: ExMachina.EctoHasManyTest."test create/2 saves overriden `has_many` associations"/1
.warning: changing assocs with change/2 or put_change/3 is deprecated, please use put_assoc/3 instead
    (stdlib) lists.erl:1262: :lists.foldl/3
    (ecto) lib/ecto/changeset.ex:247: Ecto.Changeset.change/2
    (ex_machina) lib/ex_machina/ecto.ex:128: ExMachina.Ecto.save_record/3
    test/ex_machina/ecto_has_many_test.exs:85: ExMachina.EctoHasManyTest."test create/1 creates a record without any associations"/1
.warning: changing assocs with change/2 or put_change/3 is deprecated, please use put_assoc/3 instead
    (stdlib) lists.erl:1262: :lists.foldl/3
    (ecto) lib/ecto/changeset.ex:247: Ecto.Changeset.change/2
    (ex_machina) lib/ex_machina/ecto.ex:128: ExMachina.Ecto.save_record/3
    test/ex_machina/ecto_has_many_test.exs:67: ExMachina.EctoHasManyTest."test create/1 saves `has_many` records defined in the factory"/1
.warning: changing assocs with change/2 or put_change/3 is deprecated, please use put_assoc/3 instead
    (stdlib) lists.erl:1262: :lists.foldl/3
    (ecto) lib/ecto/changeset.ex:247: Ecto.Changeset.change/2
    (ex_machina) lib/ex_machina/ecto.ex:128: ExMachina.Ecto.save_record/3
    (ex_machina) lib/ex_machina/ecto.ex:142: anonymous fn/3 in ExMachina.Ecto.persist_belongs_to_associations/2
[...]

PR #71, opened 2015-12-15
Upgrade ecto

geofflane:
Using the internal function Ecto.Association.loaded? no longer works. Fortunately there's a public function for the same thing Ecto.assoc_loaded? now

This makes #70 work (but with deprecation warnings still).

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -90,7 +90,7 @@ defmodule ExMachina.Ecto do
 
   defp not_loaded_assocs(model) do
     for {a, %{__struct__: Ecto.Association.Has}} <- get_assocs(model),
-      !Ecto.Association.loaded?(Map.get(model, a)),
+      !Ecto.assoc_loaded?(Map.get(model, a)),
       do: a
   end
 
--- a/mix.exs
+++ b/mix.exs
@@ -2,7 +2,7 @@ defmodule ExMachina.Mixfile do
   use Mix.Project
 
   @project_url "https://github.com/thoughtbot/ex_machina"
-  @version "0.5.0"
+  @version "0.6.0"
 
   def project do
     [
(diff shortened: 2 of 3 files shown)


paulcsmith:
@geofflane Thanks so much for catching this and getting started on the migration to Ecto 1.1

Do you think you could also change the other functions that are deprecated as well? If it seems like that is too much to do, then that is fine and we can do it in a separate PR. Thanks again for your work on this!

geofflane:
@paulcsmith Fixing those deprecations were much less obvious to me and I thought it might be better for someone more familiar with the code to take a look. It seems like you would need to filter out all the assocs, embeds, etc and then map them back into the changeset using `put_embed` and `put_assoc`. If that's right, I could try and take a stab at it?

paulcsmith:
@geofflane Yeah it definitely seems a bit trickier. That sounds about right. 

You'd probably have to change this function https://github.com/thoughtbot/ex_machina/blob/402e5890238e71fbff1d3285de00ca9933c33368/lib/ex_machina/ecto.ex#L81 to only get assocs (not embeds) and then add another `get_embeds` that only gets the embedded associations. If you need any help ping me and I can try to help :) If it's too tricky no worries, just be sure to leave a comment here so we can help take over

geofflane:
@paulcsmith I took a quick look, but will be honest that I probably won't have the time to do this anytime soon. So this is my official "It would be great if someone could take this" request.

paulcsmith:
@geofflane That's totally fine. I will also not have much time due to the holidays, but I'll try this PR out on a few projects and try to release it soon so it at least works with 1.1, even if there are a few deprecation warnings :)

steffenix:
Cool I just did the ecto update to 1.1 exactly what I was looking for

paulcsmith:
Thanks @geofflane! I'll get this pushed to hex in a bit

sorentwo:
The rampant warnings are making my controller tests very difficult to read. I'm looking into making the changes, hopefully a PR is forthcoming.

paulcsmith:
@sorentwo That would be wonderful. I've been a bit busy so I'd really appreciate it. If you need any help feel free to ping me!

PR #73, opened 2015-12-16
Use put_assoc/4 and put_embed/4 when building records

sorentwo:
This breaks apart the general use of `Ecto.Changeset.change` when building records. Instead, it uses `put_assoc` and `put_embed`, eliminating the deprecation warnings in Ecto 1.1.

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -93,7 +93,7 @@ defmodule ExMachina.Ecto do
 
     Enum.reduce(keys, changeset, fn(key, changes) ->
       case Map.get(record, key) do
-        %Ecto.Association.NotLoaded{} ->
+        %{__struct__: Ecto.Association.NotLoaded} ->
           changes
         association ->
           Ecto.Changeset.put_assoc(changes, key, association)
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -79,8 +79,8 @@ defmodule ExMachina.Ecto do
   end
 
   defp get_assocs(%{__struct__: struct}) do
-    for a <- struct.__schema__(:associations) do
-      {a, struct.__schema__(:association, a)}
+    for assoc <- struct.__schema__(:associations) do
+      {assoc, struct.__schema__(:association, assoc)}
     end
   end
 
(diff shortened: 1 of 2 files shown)


paulcsmith:
Really nice work. Totally agree on clarifying the naming. It all reads much better now. I'll test this on a couple projects and then merge it in!

sorentwo:
Not quite there yet. I'm testing this in a project currently and seeing a bunch of errors around invalid changesets. Pushing a fix as soon as I figure it out =)

sorentwo:
@paulcsmith Pushing the fix I have to see if it works for you as well, I'll add a failing test by tomorrow.

sorentwo:
@paulcsmith Updated with a modification that makes the association tests fail and amended the commit message with more details. Everything passes in the test suite and my project now.

test/ex_machina/ecto_has_many_test.exs
```diff
@@ -7,6 +7,7 @@ defmodule ExMachina.EctoHasManyTest do
     schema "packages" do
       field :description, :string
       has_many :statuses, ExMachina.EctoHasManyTest.PackageStatus
+      has_many :invoices, ExMachina.EctoHasManyTest.Invoice
```
paulcsmith: Could you also write a test case that makes this explicit?

Like "when a has many association is not loaded, it still saves" or something like that? And then explicitly set `invoices: Ecto.Association.NotLoaded`

That way it's clear what it's testing and we don't accidentally remove it in the future. I think this also means you could remove this from the migration and just set "statuses" to NotLoaded

lib/ex_machina/ecto.ex
```diff
@@ -79,34 +79,60 @@ defmodule ExMachina.Ecto do
...
-  defp not_loaded_assocs(model) do
-    for {a, %{__struct__: Ecto.Association.Has}} <- get_assocs(model),
-      !Ecto.assoc_loaded?(Map.get(model, a)),
-      do: a
+  defp put_assoc_changes(changeset, record) do
+    keys = assoc_keys(record) -- belongs_to_assoc_keys(record)
+
+    Enum.reduce(keys, changeset, fn(key, changes) ->
+      case Map.get(record, key) do
+        %Ecto.Association.NotLoaded{} ->
```
paulcsmith: Can you use the struct form like on line 88? That will make CI pass. Ecto is not an explicit dependency so we can't rely on it always being there

paulcsmith:
Just 2 small comments and I think we're good to go

test/ex_machina/ecto_has_many_test.exs
sorentwo: I've added an explicit case, though the simple addition of the `has_many` will make every test fail without the corresponding fix in `put_assoc_changes`.

paulcsmith: Sounds good! :+1:

sorentwo:
@paulcsmith This is all ready to go now.

paulcsmith:
Perfect. I'll try it out tomorrow. I'm at the airport now and portably won't have time to do it quite yet :) Soon though!

paulcsmith:
Sorry for the delay. Been a bit busy over the holidays but I'll try it out when I have a moment! 

Happy holidays

rranelli:
I tested the change over here and everything seems fine. 

Upon merge #70 will be good to be closed right ?

sorentwo:
Yes indeed, this closes out #70.

paulcsmith:
This looks good to me. Thanks for your work on this @sorentwo!  I'll merge it in, but I'm going to wait a bit to publish to hex. @jsteiner Before this is published, can you try this out on your app to make sure all is well? I'm a bit nervous about the association stuff since that is the trickiest thing to get right.

For anyone that wants to use this now before it's up on hex you can use the github repo in your deps

```
{:ex_machina, "~> 0.6", github: "thoughtbot/ex_machina"}
```

PR #77, opened 2015-12-22
fix typo: s/save_recod/save_record

rranelli:


paulcsmith:
@rranelli thanks and good catch!

Issue #35, opened 2015-10-09
Use do_build/2 and do_create/2 internally so that you can have custom build/create functions


paulcsmith:
I was thinking about a way to do things after/before create/build or to do intersting things with the attrs if needed. I think it would be awesome to be able to define a custom function like so

```elixir
def create(:article, attrs \\ []) do
  do_create(:article, attrs) |> with_comment
end

def build(:user, attrs \\ []) do
  email = attrs.first_name <> attrs.last_name <> "@foo.com"
  do_build(:user, Map.put_new(attrs, email: email)) 
end
```

`do_build` and `do_create` would do what `create/2` and `build/2` do today, and `create/build` would just call `do_create` and `do_build` with the passed in arguments. This would be an easier change and would allow for a very flexible build/create process

paulcsmith:
I'm open to other thoughts proposals on how to handle this though. I personally like this because it's super simple to implement and it keeps the build/create changes separate without cluttering up the main factory definition.

jsteiner:
We discussed this in person today, but I wanted to document my concern that this splits the factory definition into two places, so it's easier to miss that you're assigning the comment if you only look at the `factory` block. This may not actually matter. I need some more time to think about it, but I probably won't have time until next Friday.

paulcsmith:
Thanks for documenting that, Josh. I don't know how confusing it would be though if it's defined right underneath the factory? I definitely can see it being confusing if it's defined underneath somewhere though. Maybe we can think of something better on Friday though

josevalim:
I don't like this approach, honestly. Because the function is called `do_build`, it feels like an implementation detail. @jsteiner also brought valid concerns of splitting the factory definitions apart. Why do we need this? Why this cannot be done in the factory definition itself?

paulcsmith:
@josevalim I mostly liked overriding `build/create` because it requires very few changes to the code and seemed simple in my mind. You can do things before/after build or create without learning any new functions. Just override build/create and do whatever you need to do. 

We could do it in the factory definition though, but would require a few more functions and changing how `build` and `create` work. 

```
%Article{
  title: "My Article"
} 
|> after_create(&add_comment)

# Returns a struct like
%ExMachina.Factory{
  record: %User{title: "MyArticle"},
  before_create_functions: [],
  after_create_functions: [&add_comment]
}
```

[...]

PR #78, opened 2015-12-22
Allow to use custom changeset

tuvistavie:
Hi,
I think that changeset is quite an important part of ecto model creation,
but I could not find a way to use them with factories, so I changed the `save_record`
function to be able to use a custom changeset function.
With this change, the user can provide a `make_changeset/2` function in the factory to create
the changeset using the one defined in the model, or anywhere else.
As for `factory/1`, pattern matching can be used to choose the `make_changeset` function.
If nothing matches (e.g. `make_changeset/2` is not defined), we fallback to `Ecto.Changeset.change`.

Here is an example

```elixir
  defmodule Factory do
    use ExMachina.Ecto, repo: TestRepo

    def factory(:user) do
      %User{
        name: "John Doe",
        admin: false
      }
    end

    def factory(:article) do
      %Article{
        title: "My Awesome Article",
        author: build(:user)
      }
    end

    def make_changeset(%User{} = user, changes) do
      User.changeset(user, changes)
    end
  end

```

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -3,6 +3,8 @@ defmodule ExMachina.Ecto do
     verify_ecto_dep
     if repo = Keyword.get(opts, :repo) do
       quote do
+        @before_compile unquote(__MODULE__)
+
         use ExMachina
 
         @repo unquote(repo)
@@ -88,14 +90,14 @@ defmodule ExMachina.Ecto do
     for {key, _assoc} <- get_assocs(record), do: key
   end
 
-  defp put_assoc_changes(changeset, record) do
+  defp put_assoc_changes(changeset, record, module) do
     keys = assoc_keys(record) -- belongs_to_assoc_keys(record)
     Enum.reduce(keys, changeset, fn(key, changes) ->
       case Map.get(record, key) do
         %{__struct__: Ecto.Association.NotLoaded} ->
           changes
         associations when is_list(associations) ->
-          changesets = associations |> Enum.map(&record_to_changeset/1)
+          changesets = associations |> Enum.map(&(record_to_changeset(&1, module)))
 
           Ecto.Changeset.put_assoc(changes, key, changesets)
         association ->
@@ -153,7 +155,7 @@ defmodule ExMachina.Ecto do
     record = record |> persist_belongs_to_associations(module)
 
     record
-    |> record_to_changeset
+    |> record_to_changeset(module)
     |> repo.insert!
     |> restore_belongs_to_associations(record)
   end
@@ -190,12 +192,26 @@ defmodule ExMachina.Ecto do
     |> Map.put(association_name, association)
   end
 
-  defp record_to_changeset(%{__struct__: model, __meta__: %{__struct__: Ecto.Schema.Metadata}} = record) do
+  defp record_to_changeset(%{__struct__: model, __meta__: %{__struct__: Ecto.Schema.Metadata}} = record, module) do
     changes = record |> convert_to_changes
 
+
     struct(model)
-    |> Ecto.Changeset.change(changes)
-    |> put_assoc_changes(record)
+    |> module.make_changeset(changes)
+    |> put_assoc_changes(record, module)
     |> put_embed_changes(record)
+  end
+
+  defmacro __before_compile__(_env) do
+    # We are using line -1 because we don't want warnings coming from
+    # save_record/1 when someone defines there own save_recod/1 function.
+    quote line: -1 do
+      @doc """
+      Raises a helpful error if no factory is defined.
+      """
+      def make_changeset(record, changes) do
+        Ecto.Changeset.change(record, changes)
+      end
+    end
   end
 end
--- a/priv/test_repo/migrations/20151002190029_create_users.exs
+++ b/priv/test_repo/migrations/20151002190029_create_users.exs
@@ -5,6 +5,7 @@ defmodule ExMachina.TestRepo.Migrations.CreateUsers do
     create table(:users) do
       add :name, :string
       add :admin, :boolean
+      add :token, :string
     end
   end
 end
--- a/test/ex_machina/ecto_test.exs
+++ b/test/ex_machina/ecto_test.exs
@@ -7,6 +7,21 @@ defmodule ExMachina.EctoTest do
     schema "users" do
       field :name, :string
       field :admin, :boolean
+      field :token, :string
+    end
+
+    def changeset(model, params \\ :empty) do
+      model
+      |> cast(params, ~w(name), [])
+      |> generate_token
+    end
+
+    defp generate_token(changeset) do
+      put_change(changeset, :token, generated_token)
+    end
+
+    def generated_token do
+      "generated_token"
     end
   end
 
@@ -74,6 +89,10 @@ defmodule ExMachina.EctoTest do
         user: build(:user)
       }
     end
+
+    def make_changeset(%User{} = user, changes) do
+      User.changeset(user, changes)
+    end
   end
 
   test "raises helpful error message if no repo is provided" do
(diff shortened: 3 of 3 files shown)


paulcsmith:
I think this is an interesting idea but I'm not totally sure this is the most flexible way to do it. What about factories that aren't ecto models?

It seems like it might be better to have a way to customize the build or create functions. Then you can do whatever you need to do regardless of whether it is an ecto changeset or not. I made an issue here: https://github.com/thoughtbot/ex_machina/issues/35 for something that might work. What do you think of that? Would it work for your use case? Any thoughts on naming or inolementation?

tuvistavie:
@paulcsmith Thanks for the quick reply.
As the idea of changeset is quite specific to Ecto, I thought that something adapter specific could be more efficient here.

The use case is quite simple: I use the changeset to do what would be done in `before_create`, or some ActiveRecord callback in Rails, for example generating attributes.
What I added in the tests (generating a token for the user) is a sample usage.

As there is no standard way to deal with changesets in ex_machina, I have things like this in my factories

```elixir
  def factory(:user) do
    user = %User{
      name: sequence(:name, &("name-#{&1}"))
    }
    Map.merge(user, User.changeset(user).changes)
  end
```

This should work for most cases I think, but looks quite unintuitive for something which is very common
when using Ecto.

It gets worse when I need to choose which changeset to use, as I cannot have it in my factory anymore.
For example, in my project, I have two different changesets for my `User` model: `admin_changeset` and `changeset`.
With the current implementation, I have to do something like `build(:user) |> with_admin_changeset |> create`.
With the proposition above, I could write `create(:user, changeset: :admin)` and pattern match in `make_changeset`, which would be a lot cleaner.

About #35, I would still have to transform `map -> changeset -> map` which will get transformed to a `changeset` a moment later, but the multiple changeset case would be easier to use.
However, I do not like the idea of calling an internal function. If it were to be implemented,
I would rather see something with a signature like `build(map, map) :: map`.

paulcsmith:
@tuvistavie Those are really great examples. Thank you for sharing that. I'm still thinking about the best way to handle this. maybe allowing custom changesets is the best way, but I'm toying around with a couple other ideas. I'll post here for feedback once I've thought about it a bit more!

hunterboerner:
I've got the same problem. I need to be able to use changesets in my factories to process/validate data.

paulcsmith:
@hunterboerner Can you give some examples of what you need to process and why you need to validate the data in your factories? That would help make sure we come up with a good solution. I'm still not sure that hooking in to changesets is the best bet here, but I could be wrong. Examples will definitely help if you can share them :D

hunterboerner:
More important than validation for me is the ability to automatically set a field to a value (previously done with callbacks). I use this with my changesets:

```elixir
def changeset(model, params \\ :empty) do
  cast(...)
  |> prepare_changes(&set_build_number/1)
  ...
end
```

The field I am trying to set (build_number) is required on the database level. I can't use a sequence because the build_number is scoped to the Branch (belongs_to association). E.g.

A Build with BranchA would have build_number = 1
Another Build with BranchA would have build_number = 2
A Build with BranchB would have build_number = 1

paulcsmith:
Thank you for the example that clarifies things a lot! We may wait until Ecto 2.0 comes out since we will be changing how ExMachina.Ecto works once that happens, but I do think it makes sense to support changesets somehow. I am not sure about this implementation quite yet. I think we'll hold off a bit and wait for Ecto 2.0 since that will require a lot of changes to ExMachina.Ecto anyway.

tuvistavie:
@hunterboerner I ran in the same issue while removing callbacks, and I am not sure changeset is the best place to do this.
The rationale behind this is that `Ecto.Model` is going to disappear, leaving only `Ecto.Schema`, which seems quite close to some POEO or whatever we would call that, which is very different of what is done with activerecord patterns.
So, as soon as you start depending on the DB inside the model, which will probably happen in your example above, you break this separation of concerns.
I found that having some kind of repository pattern, or mediator pattern is much simpler to think about and to test.
I'm not sure how to integrate this properly with ex_machina though, so I'll think about it a little more and post an update here.

paulcsmith:
That would be great. Thank you @tuvistavie Ideas are always appreciated. I haven't run into this problem myself so it's a bit harder to come up with solutions :) If anyone else has examples, please feel free to post them here. The more use cases we have, the better the solution will be

tuvistavie:
@paulcsmith Here is a simplified snippet of what I am doing. https://gist.github.com/tuvistavie/8cef4edfa74da6815eb3
It is very similar to what @hunterboerner is doing, but I moved this to a service,
to avoid using the DB from my models, as explained above.

My major issue here is that I have no way of accessing the `project` from my factory,
which forces me to create a new one in it, so basically when I cannot write

```elixir
create(:build, project: project)
```

and get the expected result, as I have no way to pass this project to my changeset.
This PR would not work for this case either, but I could not find any good solution for now.
Any ideas of a good way of handling this?

Again, the `Map.merge` trick here on the `changeset.changes` really feels more like a workaround
than a long term solution.

hunterboerner:
@tuvistavie I think you misunderstand what prepare_changes/2 does. From the docs:

> Provides a function to run before emitting changes to the repository.
>
> Such function receives the changeset and must return a changeset, allowing developers to do final adjustments to the changeset or to issue data consistency commands.
>
> The given function is guaranteed to run inside the same transaction as the changeset operation for databases that do support transactions.

It does not execute set_build_number/1 until Repo.insert/update/... is called with the changeset. Your version loses the advantages of DB transactions and a developer must be aware that a specific model needs to use a _different_ changeset function in order to work properly. Could @josevalim chime in on the proper* way of doing something like this?

Thanks for the quick response on a Saturday :).

tuvistavie:
@hunterboerner @paulcsmith 
After trying different solutions, I finally came back to @hunterboerner solution of using `prepare_changes/2` and I am currently using my fork to use the changesets in the factory.
I managed to move all my `before_insert` to changesets by simply adding the relevant `def make_changeset` to my factory, so it is working for me pretty well for now.

The main issue I had without the `make_changeset` stuff was to do something like

```elixir
create(:build, project: project) # build belongs to project
```

where I needed the `project_id` to be available in my changeset to create the `build` object.

paulcsmith:
Thank you so much for reporting this and giving more details. I think allowing changesets could be a good idea. I'd like to experiment with it a bit more to see if we can make it a bit easier to setup. I'll post here with some ideas when something comes up. 

Also you could release a custom adapter on hex if you wanted. It would basically be like the current ecto adapter but have your changeset stuff. It's up to you but it might be a cool way to experiment with things.

josevalim:
It should be straight-forward to support changesets. If the factory returns a changeset, you can call `Ecto.Changeset.apply_changes` to get the struct back and then you can proceed as usual.

paulcsmith:
Thank José. That is a good point. I just wonder if that will cause issues because applying on build could mean that the project ID is not there yet so it wouldn't work correctly and only applying on create would mean for some factories build returns a struct and for other it returns a changeset. 

I'm not sure this is actually an issue but I probably won't have time for a few days to take a look. 

Either way it could be solved but I may need to change how build and create work to support both build and create, without causing unexpected results like a changeset returned by build

josevalim:
@paulcsmith ah, my solution is not a concern on Ecto 2.0. It will definitely yield different results on Ecto 1.1.

paulcsmith:
@josevalim Yeah I think I'm going to hold off on any major change to ExMachina.Ecto until Ecto 2.0 drops since I think that will make things a lot easier and will likely change how I solve most of these problems :D I'm excited for 2.0 to drop!

I still may need to change `build` slightly, but if I do need to change something I think the changes should be minor.

paulcsmith:
@tuvistavie @hunterboerner In the master branch of ExMachina we've added the ability to use custom strategies https://github.com/thoughtbot/ex_machina/blob/master/lib/ex_machina/strategy.ex

I think once #111 is closed, you could add your own strategy that allows you to do what you need. You can see an example of what this will look like in #111. Any thoughts on that?

hunterboerner:
@paulcsmith LGTM.

paulcsmith:
I'm going to close this and ping everyone when the work on strategies is finished :)

PR #79, opened 2015-12-23
Update version to 0.6.1

paulcsmith:
This version bump requires Ecto 1.1 (required since ExMachina 0.6.0). This version removes the deprecation warnings generated by Ecto 1.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,20 @@
 The noteworthy changes for each ExMachina version are included here. For a
 complete changelog, see the git history for each version via the version links.
 
-## [0.6.0] - December 16, 2015
+**To see the dates a version was published see the [hex package page].**
+
+[hex package page]: https://hex.pm/packages/ex_machina
+
+## [0.6.1]
+
+Removes warnings as reported by
+https://github.com/thoughtbot/ex_machina/issues/70. We recommend updating if you
+are using Ecto 1.1. There are no backward incompatible changes and no new
+features.
+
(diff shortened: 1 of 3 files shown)


jsteiner:
Can you update the changelog as well?

paulcsmith:
Yep I'll tackle that this week

paulcsmith:
@jsteiner Changelog added. Could you re-review when you have a sec?

CHANGELOG.md
```diff
@@ -3,6 +3,20 @@
 The noteworthy changes for each ExMachina version are included here. For a
 complete changelog, see the git history for each version via the version links.
 
+**To see the dates a version was published after 0.6.0**, see the
```
paulcsmith: It was annoying to have to add the date. Hex already does it so I figure we could just link to that page instead.

CHANGELOG.md
```diff
@@ -3,6 +3,20 @@
...
+[hex package page].
+
+[hex package page]: https://hex.pm/packages/ex_machina
+
+## [0.7.0]
```
jsteiner: I assume we're following SemVer, right? If so, this should be `0.6.1`

paulcsmith: Good point. Didn't think of that!

CHANGELOG.md
jsteiner: This qualifier is a little verbose. How about removing all the dates and simplifying this?

paulcsmith: Huge +1. I wasn’t sure about it, but I totally agree. I’ll do that

PR #80, opened 2015-12-29
Fix nested has many

rranelli:
(I have rebased this branch against the PR that bumps the version to 0.7.0)

In the current state of affairs, if we were to have factories building
nested `has_many` relationships the built models cannot be saved. When
`creating` a model with such a factory (like the shipment factory in
`ecto_has_many_test.exs`) a user would see the following error:

(see the changes in 0573f23 for the model & factory definitions)

```
** (ArgumentError) model ExMachina.EctoHasManyTest.Package has value `[%ExMachina.EctoHasManyTest.PackageStatus{__meta__: #Ecto.Schema.Metadata<:built>, id: nil, package: #Ecto.Association.NotLoaded<association :package is not loaded>, package_id: nil, status: "ordered"}, %ExMachina.EctoHasManyTest.PackageStatus{__meta__: #Ecto.Schema.Metadata<:built>, id: nil, package: #Ecto.Association.NotLoaded<association :package is not loaded>, package_id: nil, status: "sent"}, %ExMachina.EctoHasManyTest.PackageStatus{__meta__: #Ecto.Schema.Metadata<:built>, id: nil, package: #Ecto.Association.NotLoaded<association :package is not loaded>, package_id: nil, status: "shipped"}]` set for assoc named `statuses`. Assocs can only be manipulated via changesets, be it on insert, update or delete.
```

This PR fixes this issue by calling `put_assoc_changes` recursively in nested `has_many` relations.

(I would like to add more tests stressing more complex relationships if I get some more time before the review. I will probably send some commits in the next couple of days)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ In `mix.exs`, add the ExMachina dependency:
 
 ```elixir
 def deps do
-  [{:ex_machina, "~> 0.5"}]
+  [{:ex_machina, "~> 0.7"}]
 end
 ```
 
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -90,14 +90,17 @@ defmodule ExMachina.Ecto do
 
   defp put_assoc_changes(changeset, record) do
     keys = assoc_keys(record) -- belongs_to_assoc_keys(record)
-
     Enum.reduce(keys, changeset, fn(key, changes) ->
       case Map.get(record, key) do
         %{__struct__: Ecto.Association.NotLoaded} ->
           changes
+        associations when is_list(associations) ->
+          changesets = associations |> Enum.map(&record_to_changeset/1)
+
+          Ecto.Changeset.put_assoc(changes, key, changesets)
         association ->
           Ecto.Changeset.put_assoc(changes, key, association)
-        end
+      end
     end)
   end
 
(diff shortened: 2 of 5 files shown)


paulcsmith:
Hi @rranelli Thanks for this PR! I'll take a look more deeply this week. Does this also cover nested belongs_to? I think it would be a good idea to add a test case for those types of associations as well. What do you think?

rranelli:
nested `belongs_to` works just fine :)

paulcsmith:
Great! That is fantastic news. Could you add a test for nested belongs_to? If  you don't have time that is ok. LMK and I can tackle it when I have a chance

rranelli:
I have added a new commit with a test for the nested `belongs_to`.

I have noticed a somewhat unintuitive behavior I documented in the commit message (b22f4e8e).

We will probably need to address this issue in the future, but I think it does not block this PR though.

paulcsmith:
@rranelli Thank you for adding the test and documenting that behavior. I think that is ok for now. Once Ecto 2.0 comes out most of this code will be handled automatically by Ecto so it should be ok. See https://github.com/elixir-lang/ecto/issues/1102#issuecomment-166909384

paulcsmith:
Merged without the version update commits because I still need to add a changelog and clean those up a bit. Thanks for the pr @rranelli!

PR #84, opened 2016-01-08
Add Sequence.next/1 for quickly creating sequences

### LINKED ISSUE #66: Make a sequence/1 function that generates a default string
A lot of times I just want a unique value like this:

```
title: sequence(:title, fn(n) -> "Title #{n}" end)
body: sequence(:comment_body, fn(n) -> "Comment Body #{n}" end)
```

For simple cases like this I think it would be helpful if you could do something like

```
title: sequence(:title) #generates "Title #{n}"
body: sequence(:comment_body) #generates "Comment Body #{n}"
```

Thoughts on this? Would it be better to use Faker or would this be useful?

paulcsmith:
Closes #66

### DIFF
--- a/lib/ex_machina/sequence.ex
+++ b/lib/ex_machina/sequence.ex
@@ -3,6 +3,19 @@ defmodule ExMachina.Sequence do
     Agent.start_link(fn -> HashDict.new end, name: __MODULE__)
   end
 
+  def next(sequence_name) when is_binary(sequence_name) do
+    next sequence_name, fn(n)->
+      sequence_name <> " " <> to_string(n)
+    end
+  end
+
+  def next(sequence_name) do
+    raise(
+      ArgumentError,
+      "Sequence name must be a string, got #{inspect sequence_name} instead"
+    )
+  end
+
   def next(sequence_name, formatter) do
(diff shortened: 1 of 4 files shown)


lib/ex_machina/sequence.ex
```diff
@@ -3,6 +3,12 @@ defmodule ExMachina.Sequence do
     Agent.start_link(fn -> HashDict.new end, name: __MODULE__)
   end
 
+  def next(sequence_name) do
```
jsteiner: Should we raise a helpful error for non-string values?

jsteiner:
One question/comment. Otherwise :+1:

lib/ex_machina/sequence.ex
paulcsmith: That's a good point. I'm not sure though. Should we raise, or should we convert it to a string? I think raise, but just want to make sure :)

jsteiner: I think we should raise. I think anything else would be unexpected and unlikely to be helpful (if not harmful).

paulcsmith: I think you're right. I'll make it raise :)

PR #87, opened 2016-01-15
Do not recommend adding ex_machina to the applications list

josevalim:
It is mostly a test only dependency, so we can start it in the test helper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,19 +11,17 @@ In `mix.exs`, add the ExMachina dependency:
 
 ```elixir
 def deps do
-  [{:ex_machina, "~> 0.6.1"}]
+  [{:ex_machina, "~> 0.6.1", only: :test}]
 end
 ```
 
-And be sure to start the ExMachina application. For most projects (such as
-Phoenix apps) this will mean adding `:ex_machina` to the list of applications in
-`mix.exs`.
+In the snippet above, we are adding it as a test only dependency. You
+can change it accordingly if you plan to use it in other enviroments.
+Next, be sure to start the application in your `test/test_helper.exs`
+before `ExUnit.start`:
 
 ```elixir
-def application do
-  [mod: {MyApp, []},
-   applications: [:ex_machina, :other_apps...]]
-end
+{:ok, _} = Application.ensure_all_started(:ex_machina)
 ```
 
 ## Overview


paulcsmith:
@josevalim thanks for the PR! We've gone back and forth on this and it's a bit tricky. If we change the Readme to this then if they put their factory in lib/my_app/factory.ex it will fail in dev and prod because the dependency isn't there. 

In Phoenix you could put it in test/support/factory.ex, but not everyone uses Phoenix. 

We have it the way it is now because it's the least error prone that we've found so far, and because it lets you use the factory for seeds in dev/prod if you want. If we can figure out a recommendation that is simple to get started and won't cause errors once they use dev/prod environments then I'm totally up for discussing that though. 

What do you think? Also, feel free to open other issues if you have any other thoughts or suggestions. We're always looking for feedback :)

josevalim:
What if we define both modes explicitly in the README then? Test only and
all environments? I have only opened the issue because folks were running
into issues. They specified the dependency as only test and added it to
applications then their app couldn't boot!

On Sunday, January 17, 2016, Paul Smith <notifications@github.com> wrote:

> @josevalim <https://github.com/josevalim> thanks for the PR! We've gone
> back and forth on this and it's a bit tricky. If we change the Readme to
> this then if they put their factory in lib/my_app/factory.ex it will fail
> in dev and prod because the dependency isn't there.
>
> In Phoenix you could put it in test/support/factory.ex, but not everyone
> uses Phoenix.
>
> We have it the way it is now because it's the least error prone that we've
> found so far, and because it lets you use the factory for seeds in dev/prod
> if you want. If we can figure out a recommendation that is simple to get
> started and won't cause errors once they use dev/prod environments then I'm
> totally up for discussing that though.
>
> What do you think? Also, feel free to open other issues if you have any
> other thoughts or suggestions. We're always looking for feedback :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/thoughtbot/ex_machina/pull/87#issuecomment-172287283>.
>


-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

paulcsmith:
@josevalim It's definitely tricky because we had complaints when we had in test env too haha. So yeah I think including both is probably the best bet. 

If running in all environments:
  Add dep, add to applications list, put factory in lib

If running in just test
  Add test dep, start in test_helper, make sure test/support is added when in test env, and add factory to test/support

 I know you're busy. I'm happy to do this myself if you don't have time. Just LMK. :)

paulcsmith:
Also where did you hear about folks having issues? I've only seen a couple comments on an issue. I'd love to know where to look so I can catch future issues like this more quickly :)

josevalim:
Maybe we should have 3 instructions then?

* All environments
* Test environment with Phoenix
* Test environment for non Phoenix projects

That's because non Phoenix projects do not include test/support by default.

> Also where did you hear about folks having issues? I've only seen a couple comments on an issue. I'd love to know where to look so I can catch future issues like this more quickly :)

Private discussions at @plataformatec. ;)

paulcsmith:
Yeah I think 3 sections sounds best. I was thinking of keeping it just two, with a note about adding test/support, but breaking it into 3 sections probably makes it simpler. Good idea.

And thanks for bringing it to our attention! Let me know if you have any other feedback that your team has mentioned. Both good and bad helps :D

paulcsmith:
Closed in favor of #87

PR #88, opened 2016-01-19
Update installation instructions and move around "usage in a test"

paulcsmith:
@josevalim I've added the 3 separate installation instructions in #87 so I'll go ahead and close that PR.

I also changed "Where to put factories" and moved "usage in a test" closer to the overview since I think it fits in with that section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,27 @@ create_list(3, :comment, attrs)
 fields_for(:comment, attrs)
 ```
 
+## Usage in a test
+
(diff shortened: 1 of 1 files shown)


josevalim:
:shipit:

jsteiner:
:+1:

paulcsmith:
And thanks for the suggestion José :+1: Hopefully this helps people get setup however they want :D

PR #92, opened 2016-01-30
Nested association do not persist belongs_to

tokafish:
Hi there,

When using ex_machina, I noticed the following behaviour: When using a factory on a model with a `has_many` association, `belongs_to` associations of the child did not seem to be persisting on create.

I'm curious if I'm misusing the package, or if this is a genuine bug. I've added a failing spec illustrating the problem - I'll be happy to contribute a fix as well if one is warranted.

Thanks!

### DIFF
--- a/test/ex_machina/ecto_test.exs
+++ b/test/ex_machina/ecto_test.exs
@@ -7,6 +7,7 @@ defmodule ExMachina.EctoTest do
     schema "users" do
       field :name, :string
       field :admin, :boolean
+      has_many :comments, ExMachina.EctoTest.Comment
     end
   end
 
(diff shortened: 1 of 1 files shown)


paulcsmith:
@tokafish Yes I believe this is a bug. Thank you for the PR with the failing test. Once Ecto 2.0 ships I believe this will be resolved. I will keep this open so I can try it once we switch to Ecto 2.0. I'll post here with anything that comes up. Thanks again!

paulcsmith:
@tokafish It looks like this should be fixed once Ecto 2.0 drops, but for now I think this would be good to fix. If you could submit a fix for it, that would be much appreciated! If you don't have time or need help ping me and I can help out

paulcsmith:
@tokafish This should be fixed when #100 is merged. If you're using Ecto 2.0 beta then you can use ExMachina master once that PR is closed and this should fix the issue :)

Please reopen if that isn't the case

PR #98, opened 2016-02-26
Rename fields_for to params_for

paulcsmith:
This reads much more nicely IMO

```elixir
post comment_path(conn, :create), params_for(:comment)

assert Comment.changeset(%Comment{}, params_for(:comment)).valid?
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,9 +126,9 @@ create(:comment, attrs)
 create_pair(:comment, attrs)
 create_list(3, :comment, attrs)
 
-# `fields_for` returns a plain map without any Ecto specific attributes.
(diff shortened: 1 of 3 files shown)


derekprior:
fields_for sounds kind of rails-form-buildery to me. I've only ever used it for controller params or something similar. That's also how I've used `attributes_for` in FG.

paulcsmith:
Thanks, @derekprior. That brings up a point, should it be `attributes_for` like FG, or `params_for`? Personally I'm partial to params_for because it's a bit more concise and easy to type, and I think it's just as clear. What do you think?

PR #99, opened 2016-02-26
Raise if passing invalid keys to structs

### LINKED ISSUE #50: Raise a when invalid attributes are passed to `create` / `build` etc. (use struct!/2)
Elixir will raise an error if you try to pass an unknown key when creating a struct like so:

```
iex(2)> %Foo{invalid: "invalid"}
** (CompileError) iex:3: unknown key :invalid for struct Foo
```

While not every ExMachina factory generates a struct, some do. When we use `build`, we convert the passed data to a map which doesn't raise errors for invalid keys. Would it make sense to look for a `__struct__` field, and if it's there, try to convert it to a struct to raise any errors? 

Elixir 1.2 introduced [`struct!`](http://elixir-lang.org/docs/master/elixir/Kernel.html#struct!/2) which might help.

Is there any other way of doing this?

paulcsmith:
Closes #50

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ dependencies:
   pre:
     - wget http://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb && sudo dpkg -i erlang-solutions_1.0_all.deb
     - sudo apt-get update
-    - sudo apt-get install elixir=1.0.5-2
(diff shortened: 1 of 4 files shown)


BlakeWilliams:
LGTM

PR #102, opened 2016-03-04
Add insert, insert_pair and insert_list to ExMachina.Ecto

### LINKED ISSUE #89: Add functions for `insert` `insert_pair` `insert_list` in ExMachina.Ecto
These would just delegate to the `create` versions. I think this maps better to the Ecto terminology of inserting records.

`insert(:user)`
`insert_pair(:user)`
`insert_list(:user)`

Thoughts?

paulcsmith:
This is totally overhauled from what we were originally going to do. Instead base ExMachina will only add `build` functions. You can use `ExMachina.Strategy` to define more functions. ExMachina.Ecto uses this under the hood to add `insert` functions to factories.

This PR also removes a bunch of old stuff we're no longer using

Closes #89


lib/ex_machina.ex
```diff
@@ -5,7 +5,7 @@ defmodule ExMachina do
   In depth examples are in the [README](README.html)
   """
 
-  defmodule UndefinedFactory do
+  defmodule UndefinedFactoryError do
```
paulcsmith: It is Elixir convention to add `Error` to all errors

README.md
```diff
@@ -126,6 +126,12 @@ create(:comment, attrs)
 create_pair(:comment, attrs)
 create_list(3, :comment, attrs)
 
+# `insert*` returns an inserted comment. Only works with ExMachina.Ecto
+# Works the same as create, but keeps in line with Ecto terminology.
+create(:comment, attrs)
```
jsteiner: These should be `insert`

paulcsmith: Oops :)

README.md
```diff
@@ -206,22 +207,7 @@ end
...
-`ModelCase`  modules (found in `test/support/conn_case.ex` and
-`test/support/model_case.ex`). To import your factory we recommend excluding
-`build/2` or aliasing your factory instead.
-
-```elixir
-# in test/support/conn_case|model_case.ex
-
-# Add `except: [build: 2] to the `Ecto.Model` import
-import Ecto.Model, except: [build: 2]
-```
```
paulcsmith: Ecto changed the function to `Ecto.build_assoc` so this no longer applies

README.md
```diff
@@ -231,51 +217,33 @@ change this line in `mix.exs`:
...
 ```
 
-## Using without Ecto
+## Custom strategies (using ExMachina without Ecto)
 
-You can use ExMachina without Ecto, by using just the `build` function, or by
-defining `save_record/1` in your module.
+You can use ExMachina without Ecto, by using just the `build` functions, or you
+can define one or more custom strategies to use in your factory. Here's an
+example, but check out the docs on ExMachina.Strategy for more info.
```
paulcsmith: I need to link to this once published to hexdocs

paulcsmith:
I'll fix the error on Circle sometime this week. @jsteiner any other comments?

README.md
jsteiner: Awesome!

README.md
jsteiner: Can we predict what it is going to be? Just want to make sure we don't forget

README.md
```diff
@@ -231,51 +217,33 @@ change this line in `mix.exs`:
...
-```
-
-You can do something similar while also using Ecto by defining a new function.
-This gives you the power to call `create` and save to Ecto, or call `build_json`
-or `create_json` to return encoded JSON objects.
-
-```elixir
-defmodule MyApp.Factory do
-  use ExMachina.Ecto, repo: MyApp.Repo
+defmodule MyApp.JsonFactory do
```
jsteiner: Can we just call this `MyApp.Factory`? I think the `JsonFactory` calls itself out a little too much. Took me a couple seconds to realize it wasn't actually doing any magic and it's irrelevant.

README.md
jsteiner: Also, `ExMachina.EctoStrategy` is a pretty good example. Maybe link there too?

jsteiner:
Nice! I had a few small comments, but otherwise I think this is a really elegant solution. Great work @paulcsmith!

paulcsmith:
Thanks for the review @jsteiner! I'm glad you like the new direction. I'll resolve the comments and get this in on Friday :)

README.md
paulcsmith: Great idea!

PR #110, opened 2016-03-26
Use new factory definition syntax

### LINKED ISSUE #101: Proposal to change factory definition
The factory definition syntax right now uses the same function name, but pattern matches on the first argument. This works ok, but it does cause a few problems

* You can't easily use ctags/symbol view to jump to factories because they are all the same function
* You can't add functions that are used by an factory in between other factories since they are all the same function and elixir will raise a warning.

I propose that we change the factory definition syntax to {factory_name}_factory. This will also make it so we can add callbacks that are grouped near the factory in the future

```elixir
defmodule Factory do
  # This is nice because you can add functions for the user factory near it instead of at the bottom
  # Also works better with symbol view so it doesn't show a bunch of `factory` functions.
  def user_factory do
    %User{
      first_name: "Paul",
      last_name: "Smith",
    }
  end
  
  # You can now define helper functions near the factory that it applies to
  def make_admin(user) do
    assing_to_admin_group(user)
  end
  
  # In the future we could add callbacks like this. If we keep the current syntax you'd have to put these
  # at the bottom of the file which makes it hard to see what a factory can do
  def after_build_user(user, attrs) do
    email = attrs.first_name <> attrs.last_name <> "@foo.com"
    Map.put_new(user, :email, email)
  end
  
  def comment_factory do
    %Comment{
      body: "This is my comment"
    }
  end
end
```

You would still call these with the same function `build(:user)`, etc.

paulcsmith:
Pros

* Can use symbol view/ctags to jump to the factories more easily
* Can add helper functions underneath the factory they belong to
* IMO it's easier to scan your factory module for a particular factory
* Easier to type :)
* If/when callbacks are introduced, this will allow you to put the
  callback functions near the factory

Cons

* A little more *magic* since you still call the functions with atoms,
  e.g.`build(:user)`

Closes #101

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,14 +83,14 @@ defmodule MyApp.Factory do
   # without Ecto
   use ExMachina
 
-  def factory(:user) do
+  def user_factory do
     %User{
       name: "Jane Smith",
       email: sequence(:email, &"email-#{&1}@example.com"),
     }
   end
 
-  def factory(:article) do
+  def article_factory do
     %Article{
       title: "Use ExMachina!",
       # associations are inserted when you call `create`
@@ -99,7 +99,7 @@ defmodule MyApp.Factory do
     }
   end
 
-  def factory(:comment) do
+  def comment_factory do
     %Comment{
       text: "It's great!",
       article: build(:article),
@@ -179,7 +179,7 @@ and embeds. Since we automatically save these records for you, we advise that
 factory definitions only use `build/2` when declaring associations, like so:
 
 ```elixir
-def factory(:article) do
+def article_factory do
   %Article{
     title: "Use ExMachina!",
     # associations are inserted when you call `create`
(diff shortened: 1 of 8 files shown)


jsteiner:
I'm down with this, mainly for defining helper functions and callbacks near the function.

I'm not sure how often ctags would be useful since it would be rare to use the function directly.

paulcsmith:
Good point. Ctags are probably useless here, but a lot of editors have "Jump to Definition/Symbol Views" where you can jump to a function. Right now they're all called `factory` so this should make it easier for people using those editors (I hope :D)

I'll go ahead and merge this in. Thanks for the review!

quolpr:
The regex that can help you to migrate to new factory version:
Replace `def factory\(:(.*?)\)` with `def $1`

Can be done in atom/vscode

PR #114, opened 2016-04-05
Include note about README being out of date

### LINKED ISSUE #113: Mismatch between readme usage instruction in master and current release version
Hi,
I tried to work by the instructions. If I am correct, the readme refers to the new way of declaring factory methods, and the current version specified by the readme - 0.6.1 - does not yet include that code.

Readme:
```elixir
 def user_factory do
    %User{
      name: "Jane Smith",
      email: sequence(:email, &"email-#{&1}@example.com"),
    }
  end

build(:comment, attrs)
```

Current release code in 0.6.1 ex_machina.ex:
```elixir
def build(module, factory_name, attrs \\ %{}) do
    attrs = Enum.into(attrs, %{})
    module.factory(factory_name) |> Map.merge(attrs)
  end
```

New code in ex_machina.ex:
https://github.com/thoughtbot/ex_machina/blob/master/lib/ex_machina.ex#L98
```elixir
def build(module, factory_name, attrs \\ %{}) do
    attrs = Enum.into(attrs, %{})
    function_name = Atom.to_string(factory_name) <> "_factory" |> String.to_atom
    if Code.ensure_loaded?(module) && function_exported?(module, function_name, 0) do
      apply(module, function_name, []) |> do_merge(attrs)
    else
      raise UndefinedFactoryError, factory_name
    end
  end
```

paulcsmith:
Closes #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,9 @@
 ExMachina makes it easy to create test data and associations. It works great
 with Ecto, but is configurable to work with any persistence library.
 
+> **This README is for the unreleased beta version of ExMachina**. Here are the 
+[docs for the latest published version of ExMachina](https://hexdocs.pm/ex_machina/README.html).
+
 ## Installation
 
 #### To install in all environments (useful for generating seed data in dev/prod):
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 ExMachina makes it easy to create test data and associations. It works great
 with Ecto, but is configurable to work with any persistence library.
 
-> **This README is for the unreleased beta version of ExMachina**. Here are the 
+> **This README follows master, which may not be the currently published version**. Here are the 
 [docs for the latest published version of ExMachina](https://hexdocs.pm/ex_machina/README.html).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,9 @@
 ExMachina makes it easy to create test data and associations. It works great
 with Ecto, but is configurable to work with any persistence library.
 
+> **This README is for the unreleased beta version of ExMachina**. Here are the
```
drapergeek: How about something like:

This README follows master, which may not be the currently released version.

Not picky about the wording. My main concerns were to make it clear that sometimes the docs may be the same and also that this isn't necessarily a "beta" version since it's just master.

paulcsmith: In this case it is for a beta version, but you are right that that probably isn’t as clear. I like your suggested wording. I’ll go with that (or something very close to it)

drapergeek:
👍

paulcsmith:
Thanks for the review!

PR #115, opened 2016-04-12
Omit trailing space from sequence shortcut

### LINKED ISSUE #108: Remove space from the end when using sequence with a string
This will make it so you can use it for things like usernames more easily

sequence("username")

paulcsmith:
Closes #108

This makes it so the sequence short function can be used for things like
usernames where a space will make the username invalid.


drapergeek:
LGTM

paulcsmith:
Thanks for the review!

PR #119, opened 2016-04-18
Update README examples for defining factory to newer syntax

amorphid:
Update README to change factory examples from `def foo_factory do` to `def factory(:foo) do`.  The older syntax no longer works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 ExMachina makes it easy to create test data and associations. It works great
 with Ecto, but is configurable to work with any persistence library.
 
-> **This README follows master, which may not be the currently published version**. Here are the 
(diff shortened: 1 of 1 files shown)


paulcsmith:
Thank you for tackling this! However, the syntax in the README is for the unreleased beta version of ExMachina, and is correct. I tried to add a note to the top that made that clear, but maybe I can add an additional note in the section for defining factories, so people realize this is for master. If you go to the linked docs (https://hexdocs.pm/ex_machina/README.html) you will see that it has the correct README for the currently released version. I'm waiting to publish the latest version of ExMachina when Ecto 2.0 comes out of beta, since ExMachina 1.0 will require Ecto 2.0 to work

amorphid:
Ah, gotcha.  Coolio.

PR #122, opened 2016-04-29
Replace 'create' for 'insert' on README

fladson:


paulcsmith:
These README and doc improvements are much appreciated since it will reduce confusion for newcomers. Thank you very much, and good catch!

PR #123, opened 2016-05-05
Don't return id field in params_for

jsteiner:
I can't think of any use case where you would want the id field when
using params_for. For something scenarios like submitting to an API, or
getting the bare fields to test validations it would be rare to
actually submit the id itself, and so the current behavior feels
unexpected.

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -50,11 +50,11 @@ defmodule ExMachina.Ecto do
   This is only for use with Ecto models.
 
   Will return a map with the fields and virtual fields, but without the Ecto
-  metadata and associations.
+  metadata, associations, and the primary key.
 
   ## Example
 
-      def factory(:user) do
+      def user_factory do
         %MyApp.User{name: "John Doe", admin: false}
       end
 
@@ -71,6 +71,7 @@ defmodule ExMachina.Ecto do
     |> Map.from_struct
     |> Map.delete(:__meta__)
     |> Map.drop(struct.__schema__(:associations))
+    |> Map.drop(struct.__schema__(:primary_key))
   end
   defp drop_ecto_fields(record) do
     raise ArgumentError, "#{inspect record} is not an Ecto model. Use `build` instead."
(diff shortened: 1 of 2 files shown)


paulcsmith:
I like the idea of this, but I think we should delete the record's primary key instead of assuming it is :id. 

You can get this with the functions that Ecto defines for schemas

```elixir
MyModel.__schema__(:primary_key)
```

paulcsmith:
Also we should document that the primary key is removed

lib/ex_machina/ecto.ex
```diff
@@ -69,10 +69,13 @@ defmodule ExMachina.Ecto do
...
-    |> Map.drop(struct.__schema__(:associations))
+    |> Map.drop(ecto_fields(struct))
   end
   defp drop_ecto_fields(record) do
     raise ArgumentError, "#{inspect record} is not an Ecto model. Use `build` instead."
   end
+
+  defp ecto_fields(struct) do
+    [:__meta__] ++ struct.__schema__(:associations) ++ struct.__schema__(:primary_key)
+  end
```
paulcsmith: I think this feels a bit roundabout because we already knowing it's dropping ecto fields because this function is called from `drop_ecto_fields`. So I think the abstraction is probably not necessary. What do you think about

```elixir
record
|> Map.from_struct
|> Map.delete(:__meta__)
|> Map.delete(struct.schema__(:associations))
|> Map.delete(struct.schema__(:primary_key))
```

However, I don't feel strongly so if you do, feel free to leave it as is

lib/ex_machina/ecto.ex
```diff
@@ -50,7 +50,7 @@ defmodule ExMachina.Ecto do
   This is only for use with Ecto models.
 
   Will return a map with the fields and virtual fields, but without the Ecto
-  metadata and associations.
+  metadata, associations, or the primary key.
```
paulcsmith: I think "and" makes more sense here. Otherwise it seems like it might delete metadata, or it might delete associations or it might drop primary key. In reality it drops them all

paulcsmith:
Two small comments and then LGTM :D

lib/ex_machina/ecto.ex
jsteiner: That's fair.

Issue #105, opened 2016-03-18
Add params_with_assocs/2 to ExMachina.Ecto


paulcsmith:
Based on the issue by @geofflane in #104. I would love help on this if someone has time, if not I hope to tackle this in a few weeks.

* Add `params_with_assocs/2` to ExMachina.Ecto
* Should accept the same arguments as `ExMachina.Ecto.params_for/2`
* It will insert `belongs_to` relationships from the factory, but *not* insert the factory itself. e.g. if you call `params_with_assocs(:comment)` it should insert the comment's belongs_to associations, but *not* the comment itself.
* After inserting the associations it should set the  association's id. So if a `Comment` belongs_to a `Post`, `params_with_assocs(:comment)` would insert the post and use the post's primary key and put on on the comment.
* It should also strip out the Ecto associations and other Ecto fields. This will work the same as `params_for/2` works currently. https://github.com/thoughtbot/ex_machina/blob/f6285d2033933424d78870e70810acae8885e1fb/lib/ex_machina/ecto.ex#L75
* You should be able to override associations, e.g. `params_with_assocs(:comment, post: build(:post))`. If the post is built then it will insert it and set the appropriate foreign key. If the association has already been inserted it will not try to insert it again, and will set the appropriate foreign key.

jsteiner:
Done in https://github.com/thoughtbot/ex_machina/pull/124

PR #124, opened 2016-05-05
Add params_with_assocs function

jsteiner:
Addresses: https://github.com/thoughtbot/ex_machina/issues/105

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -23,6 +23,10 @@ defmodule ExMachina.Ecto do
           ExMachina.Ecto.params_for(__MODULE__, factory_name, attrs)
         end
 
+        def params_with_assocs(factory_name, attrs \\ %{}) do
+          ExMachina.Ecto.params_with_assocs(__MODULE__, factory_name, attrs)
+        end
+
         def fields_for(factory_name, attrs \\ %{}) do
           raise "fields_for/2 has been renamed to params_for/2."
         end
@@ -52,6 +56,9 @@ defmodule ExMachina.Ecto do
   Will return a map with the fields and virtual fields, but without the Ecto
   metadata, associations, and the primary key.
 
+  If you want belongs_to associations to be inserted, use
+  `params_with_assocs/2`.
+
   ## Example
 
       def user_factory do
@@ -64,6 +71,49 @@ defmodule ExMachina.Ecto do
   def params_for(module, factory_name, attrs \\ %{}) do
     module.build(factory_name, attrs)
     |> drop_ecto_fields
+  end
+
+  @doc """
+  Same as `params_for/2`, but inserts all belongs_to associations and sets the
+  foreign keys.
+
+  ## Example
+
+      def article_factory do
+        %MyApp.Article{title: "An Awesome Article", author: build(:author)}
+      end
+
+      # Inserts an author and returns %{title: "An Awesome Article", author_id: 12}
+      params_with_assocs(:article)
+  """
+  def params_with_assocs(module, factory_name, attrs \\ %{}) do
+    module.build(factory_name, attrs)
+    |> insert_belongs_to_assocs(module)
+    |> drop_ecto_fields
+  end
+
+  defp insert_belongs_to_assocs(record = %{__struct__: struct, __meta__: %{__struct__: Ecto.Schema.Metadata}}, module) do
+    Enum.reduce(struct.__schema__(:associations), record, fn(association_name, record) ->
+      case struct.__schema__(:association, association_name) do
+        association = %{__struct__: Ecto.Association.BelongsTo} ->
+          insert_built_belongs_to_assoc(
+            module,
+            association.owner_key,
+            association_name,
+            record
+          )
+        _ -> record
+      end
+    end)
+  end
+
+  def insert_built_belongs_to_assoc(module, foreign_key, association_name, record) do
+    case Map.get(record, association_name) do
+      built_relation = %{__meta__: %{state: :built}} ->
+        relation = built_relation |> module.insert
+        Map.put(record, foreign_key, relation.id)
+      _ -> Map.delete(record, foreign_key)
+    end
   end
 
   defp drop_ecto_fields(record = %{__struct__: struct, __meta__: %{__struct__: Ecto.Schema.Metadata}}) do
--- a/test/ex_machina/ecto_test.exs
+++ b/test/ex_machina/ecto_test.exs
@@ -42,4 +42,35 @@ defmodule ExMachina.EctoTest do
       TestFactory.params_for(:user_map)
     end
   end
+
+  test "params_with_assocs/2 inserts belongs_tos that are set by the factory" do
+    assert has_association_in_schema?(ExMachina.Article, :editor)
+
+    assert TestFactory.params_with_assocs(:article) == %{
+      title: "My Awesome Article",
+      author_id: ExMachina.TestRepo.first!(User).id,
+    }
+  end
+
+  test "params_with_assocs/2 doesn't insert unloaded assocs" do
+    not_loaded = %{__struct__: Ecto.Association.NotLoaded}
+
+    assert TestFactory.params_with_assocs(:article, editor: not_loaded) == %{
+      title: "My Awesome Article",
+      author_id: ExMachina.TestRepo.first!(User).id,
+    }
+  end
+
+  test "params_with_assocs/2 doesn't try to save has_many fields" do
+    assert has_association_in_schema?(ExMachina.User, :articles)
+
+    assert TestFactory.params_with_assocs(:user) == %{
+      admin: false,
+      name: "John Doe",
+    }
+  end
+
+  defp has_association_in_schema?(model, association_name) do
+    Enum.member?(model.__schema__(:associations), association_name)
+  end
 end
--- a/test/support/models/article.ex
+++ b/test/support/models/article.ex
@@ -3,6 +3,8 @@ defmodule ExMachina.Article do
 
   schema "articles" do
     field :title, :string
+
     belongs_to :author, ExMachina.User
+    belongs_to :editor, ExMachina.User
   end
 end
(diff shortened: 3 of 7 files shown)


lib/ex_machina/ecto.ex
```diff
@@ -66,6 +70,33 @@ defmodule ExMachina.Ecto do
     |> drop_ecto_fields
   end
 
+  @doc """
+  Same as `params_for/2`, but it inserts all associations and sets the
+  foreign_keys.
```
paulcsmith: I think that it should only insert the `belongs_to` associations and not `has_many`. What do you think?

`Same as `params_for/2`, but it inserts all belongs_to associations and sets the foreign keys`

lib/ex_machina/ecto.ex
```diff
@@ -66,6 +70,33 @@ defmodule ExMachina.Ecto do
...
+      def article_factory do
+        %MyApp.Article{title: "An Awesome Article", author: build(:user)}
+      end
+
+      # Returns %{title: "An Awesome Article", author_id: 12}
+      params_with_assocs(:article)
+  """
+  def params_with_assocs(module, factory_name, attrs \\ %{}) do
+    module.build(factory_name, attrs)
+    |> insert_ecto_fields(module)
```
paulcsmith: Maybe `insert_belongs_to_associations`?

lib/ex_machina/ecto.ex
```diff
@@ -66,6 +70,34 @@ defmodule ExMachina.Ecto do
...
+    |> drop_ecto_fields
+  end
+
+  defp insert_ecto_fields(record = %{__struct__: struct, __meta__: %{__struct__: Ecto.Schema.Metadata}}, module) do
+    Enum.reduce(struct.__schema__(:associations), record, fn(association, record) ->
+      foreign_key = struct.__schema__(:association, association).owner_key
+      relation = Map.get(record, association) |> module.insert
```
jsteiner: This is going to break for `has_many`s. Fixing.

test/ex_machina/ecto_test.exs
```diff
@@ -67,4 +82,17 @@ defmodule ExMachina.EctoTest do
...
+    params = Factory.params_with_assocs(:article)
+
+    assert Map.keys(params) == [:author_id, :title]
+
+    %{
+      title: "My Awesome Article",
+      author_id: author_id,
+    } = params
+
+    assert author_id != nil
```
paulcsmith: I think this might be easier to visualize if we asserted against an exact map

```elixir
assert params == %{
  title: "My Awesome Article",
  author_id: TestRepo.one!(User).id
}
```

jsteiner: Cool. I dig it.

test/ex_machina/ecto_test.exs
```diff
@@ -10,13 +10,28 @@ defmodule ExMachina.EctoTest do
     end
   end
 
+  defmodule Article do
+    use Ecto.Schema
+    schema "articles" do
+      field :title, :string
+      belongs_to :author, User
```
paulcsmith: I think it would also be good to add another `belongs_to` that isn't defined in the factory to make sure there are no issues when the association is not set

lib/ex_machina/ecto.ex
```diff
@@ -66,6 +70,34 @@ defmodule ExMachina.Ecto do
...
   end
 
+  @doc """
+  Same as `params_for/2`, but it inserts all associations and sets the
+  foreign_keys.
+
+  ## Example
+
+      def article_factory do
+        %MyApp.Article{title: "An Awesome Article", author: build(:user)}
```
paulcsmith: Maybe `build(:author)` so the terminology remains the same throughout the example?

lib/ex_machina/ecto.ex
```diff
@@ -66,6 +70,34 @@ defmodule ExMachina.Ecto do
...
+      end
+
+      # Inserts an author and returns
+      # %{title: "An Awesome Article", author_id: 12}
+      params_with_assocs(:article)
```
paulcsmith: I like this example a lot

lib/ex_machina/ecto.ex
```diff
@@ -66,6 +70,34 @@ defmodule ExMachina.Ecto do
...
+  """
```
paulcsmith: It might be worth adding a note to `params_for/2` that belongs_to associations are not included and they should look at `params_with_assocs/2` if they need it. Otherwise I'm not sure people will realize it's here or why it matters

paulcsmith:
This is awesome! I was actually going to ask if you had time to tackle this next :D

lib/ex_machina/ecto.ex
```diff
@@ -66,6 +73,50 @@ defmodule ExMachina.Ecto do
     |> drop_ecto_fields
   end
 
+  @doc """
+  Same as `params_for/2`, but it inserts all belongs_to associations and sets the
```
paulcsmith: Maybe this could be shortened to `Same as `params_for/2`, but inserts all belongs_to associations and sets the foreign keys`? If you don't like that I'm fine with it as-is as well

lib/ex_machina/ecto.ex
```diff
@@ -66,6 +73,50 @@ defmodule ExMachina.Ecto do
...
+  foreign_keys.
+
+  ## Example
+
+      def article_factory do
+        %MyApp.Article{title: "An Awesome Article", author: build(:author)}
+      end
+
+      # Inserts an author and returns
+      # %{title: "An Awesome Article", author_id: 12}
```
paulcsmith: This could probably fit on one line and might read better that way

lib/ex_machina/ecto.ex
```diff
@@ -66,6 +73,50 @@ defmodule ExMachina.Ecto do
...
+            association.owner_key,
+            association_name,
+            record
+          )
+        _ -> record
+      end
+    end)
+  end
+
+  def insert_built_belongs_to_assoc(module, foreign_key, association_name, record) do
```
paulcsmith: maybe instead of `module` it could be `repo`?

paulcsmith: And maybe that should also be changed for the other functions. Might be better for a separate PR though. I'd be fine with this as-is

test/ex_machina/ecto_test.exs
```diff
@@ -42,4 +42,18 @@ defmodule ExMachina.EctoTest do
...
   end
+
+  test "params_with_assocs/2 inserts belongs_to associations" do
+    assert TestFactory.params_with_assocs(:article) == %{
+      title: "My Awesome Article",
+      author_id: ExMachina.TestRepo.first!(User).id,
+    }
+  end
+
+  test "params_with_assocs/2 inserts associated fields" do
```
paulcsmith: I'm not sure I understand this test. What is this covering that the previous one isn't?

test/support/models/article.ex
```diff
@@ -3,6 +3,8 @@ defmodule ExMachina.Article do
 
   schema "articles" do
     field :title, :string
+
     belongs_to :author, ExMachina.User
+    belongs_to :editor, ExMachina.User
```
paulcsmith: I think a comment explaining why this is here might be good. Usually I don't like comments, but without that I think it is likely someone will delete this and wonder why all the tests still pass. What do you think?

paulcsmith:
A few comments, but this looks really good :) Good to merge whenever you're ready

lib/ex_machina/ecto.ex
jsteiner: Agreed that this would be better, but should probably be part of another PR.

test/support/models/article.ex
jsteiner: Thoughts on this, instead? https://github.com/thoughtbot/ex_machina/pull/124/commits/9a0cc9dbeafe4d7a19d080d01dccf60ad2003572

test/ex_machina/ecto_test.exs
```diff
@@ -43,14 +43,18 @@ defmodule ExMachina.EctoTest do
     end
   end
 
-  test "params_with_assocs/2 inserts belongs_to associations" do
+  test "params_with_assocs/2 inserts belongs_tos that are set by the factory" do
+    assert Enum.member?(ExMachina.Article.__schema__(:associations), :editor)
```
paulcsmith: I like this, but I think extracting to a private function would be good. Something like `assert has_association_in_schema?(ExMachina.Article, :editor)`. What do you think?

test/ex_machina/ecto_test.exs
```diff
@@ -42,4 +42,22 @@ defmodule ExMachina.EctoTest do
       TestFactory.params_for(:user_map)
     end
   end
+
+  test "params_with_assocs/2 inserts belongs_tos that are set by the factory" do
+    assert Enum.member?(ExMachina.Article.__schema__(:associations), :editor)
```
paulcsmith: Reposting because the comment was buried for some reason:

> I like this, but I think extracting to a private function would be good. Something like assert has_association_in_schema?(ExMachina.Article, :editor). What do you think?

Also, I'm not exactly sure what "editor" looks like when it's not set. So it might be nice to have an assertion for that. That would also ensure that someone doesn't accidentally add it to the factory and remove test coverage

jsteiner: What do you mean by `I'm not exactly sure what "editor" looks like when it's not set`? If somebody adds `:editor` to the factory this test will fail because it will be included in the hash returned by `params_with_assocs`.

paulcsmith: Sorry I wasn't that clear about that. I meant that this test relies on `editor` being something, but it's not clear in this test what it is. You'd have to look at the schema to realize that it's defined, but not being set in the factory. So there is no way to look at this test and realize that `editor` is important in any way. So I think that you could assert that it's on the schema and give it an explanatory name like `defined_on_schema?(Article, :editor)` or something

paulcsmith: So I guess what I'm saying is that editor is a magic value. It's not clear that it's being tested

PR #125, opened 2016-05-06
Fix undefined struct error

lukerollans:
If an Elixir / Phoenix novice (like me) follows these instructions literally, structs defined in `factory.ex` will not be defined correctly

This updates `README.md` to include the required `alias` statements in context

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,10 @@ defmodule MyApp.Factory do
 
   # without Ecto
   use ExMachina
+  
+  alias MyApp.User
+  alias MyApp.Article
+  alias MyApp.Comment
 
   def user_factory do
     %User{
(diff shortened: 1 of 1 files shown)


paulcsmith:
@lukerollans This is a very good point! I think aliasing still might be easy to miss, and isn't what I'd necessarily recommend in factories since the alias is often used just once (for defining the factory)

What do you think of removing the aliases and doing this for all the factories?

```elixir
def user_factory do
  %MyApp.User{
  }
end
```

That's more in line with what I would recommend doing, and is easier to notice I think.

lukerollans:
I'm usually a fan of aliasing, but I suppose that given in this context where the struct name only needs to be used once, you suggestion does make more sense. And you're right, makes it more noticable

PR #127, opened 2016-05-06
Drop the database for each test run

jsteiner:
Now that we are using a single file for migrations, we should drop the
database each test run to get the new schema.


paulcsmith:
LGTM

PR #128, opened 2016-05-06
Fix #90 Insert Twice

### LINKED ISSUE #90: Raise helpful error if you accidentally call insert twice
I accidentally did this:

```elixir
appointment_booking = create(:appointment_booking)
  |> set_appointment_price(500)
  |> insert
```

And got this error:

```
* (Ecto.ConstraintError) constraint error when attempting to insert model:

         * unique: appointment_booking_pkey

     If you would like to convert this constraint into an error, please
     call unique_constraint/3 in your changeset and define the proper
     constraint name. The changeset has not defined any constraint.
```

I think it would be helpful if it said something like:

```
you called `insert` on #{the record} but it was already saved. You can only call insert on records that haven't been inserted yet
```

Or something along those lines. This is pretty low priority, but might be nice to have

davidkuhta:
Hey there, so I think this one is good to go.

Note, I split handle_insert function per @paulcsmith comment in #90 
```
  def handle_insert(%{__meta__: %{state: :loaded}} = record, _) do
    raise ArgumentError,
    "you called `insert` on #{inspect record} but it has already been saved.
     You can only call `insert` on records that haven't been inserted yet"
  end

  def handle_insert(%{__meta__: %{__struct__: Ecto.Schema.Metadata}} = record, %{repo: repo}) do
    repo.insert!(record)
  end
```

Note: I rebased with the latest test changes from @jsteiner regarding Factory -> TestFactory

Let me know if it needs anything else. Cheers!

### DIFF
--- a/lib/ex_machina/ecto_strategy.ex
+++ b/lib/ex_machina/ecto_strategy.ex
@@ -3,6 +3,15 @@ defmodule ExMachina.EctoStrategy do
 
   use ExMachina.Strategy, function_name: :insert
 
+  def handle_insert(%{__meta__: %{state: :loaded}} = record, _) do
+    raise "You called `insert` on a record that has already been inserted.
+     Make sure that you have not accidentally called insert twice.
+
+     The record you attempted to insert:
+
+     #{inspect record, limit: :infinity}"
+  end
+
   def handle_insert(%{__meta__: %{__struct__: Ecto.Schema.Metadata}} = record, %{repo: repo}) do
     repo.insert! record
   end
--- a/test/ex_machina/ecto_strategy_test.exs
+++ b/test/ex_machina/ecto_strategy_test.exs
@@ -46,4 +46,11 @@ defmodule ExMachina.EctoStrategyTest do
 
     assert article.author == my_user
   end
+
+  test "insert/1 raises if attempting to insert already inserted record" do
+    message = ~r/You called `insert` on a record that has already been inserted./
+    assert_raise RuntimeError, message, fn ->
+      TestFactory.insert(:user, name: "Maximus") |> TestFactory.insert
+    end
+  end
 end


lib/ex_machina/ecto_strategy.ex
```diff
@@ -3,8 +3,14 @@ defmodule ExMachina.EctoStrategy do
 
   use ExMachina.Strategy, function_name: :insert
 
+  def handle_insert(%{__meta__: %{state: :loaded}} = record, _) do
+    raise ArgumentError,
+    "you called `insert` on #{inspect record} but it has already been saved.
```
paulcsmith: So I just realized this may not be ideal since the record could be *really big* so maybe something like

```elixir
you called `insert` on a record that has already been inserted # Note the wording changed to "been inserted"
You can only call `insert` on records that haven't been inserted yet.

The record you attempted to save:

#{inspect record, limit: :infinity}
```

That way the record comes afterward and doesn't appear in the middle of the sentence

test/ex_machina/ecto_strategy_test.exs
```diff
@@ -46,4 +46,12 @@ defmodule ExMachina.EctoStrategyTest do
 
     assert article.author == my_user
   end
+
+  test "insert/1 raises if attempting to insert an existing model" do
+    message = ~r/You can only call `insert` on records that haven't been inserted yet/
+    assert_raise ArgumentError, message, fn ->
+      TestFactory.insert(:user, name: "Maximus")
```
paulcsmith: I would put the pipe on the same line

test/ex_machina/ecto_strategy_test.exs
```diff
@@ -46,4 +46,12 @@ defmodule ExMachina.EctoStrategyTest do
 
     assert article.author == my_user
   end
+
+  test "insert/1 raises if attempting to insert an existing model" do
```
paulcsmith: How about "if attempting to insert already inserted record"?

paulcsmith:
Add a few comments and then this looks good to merge. Nice work and thank you so much!

davidkuhta:
Haha that was quick, I didn't realize you were reviewing, as I was just squashing the commits.

lib/ex_machina/ecto_strategy.ex
```diff
@@ -3,10 +3,20 @@ defmodule ExMachina.EctoStrategy do
 
   use ExMachina.Strategy, function_name: :insert
 
-  def handle_insert(%{__meta__: %{__struct__: Ecto.Schema.Metadata}} = record, %{repo: repo}) do
-    repo.insert! record
+  def handle_insert(%{__meta__: %{state: :loaded}} = record, _) do
+    raise ArgumentError,
```
paulcsmith: I think `raise` without `ArgumentError` is ok. A `RuntimeError` should be fine here

lib/ex_machina/ecto_strategy.ex
```diff
@@ -3,10 +3,20 @@ defmodule ExMachina.EctoStrategy do
...
+    "you called `insert` on a record that has already been inserted.
+     You can only call `insert` on records that haven't been inserted yet.
```
paulcsmith: Now that I'm reading this again it seems a bit redundant. Maybe the second sentence should be "Make sure that you have not accidentally called `insert` twice." What do you think?

davidkuhta: 👍  Agreed

paulcsmith:
Thanks @davidkuhta! 🎉

PR #133, opened 2016-05-17
Renamed factory.ex to test_factory.ex

davidkuhta:
Renamed `test/support/factory.ex` to `test/support/test_factory.ex` consistent with module name: `ExMachina.TestFactory`


paulcsmith:
Looks good! Thank you!

PR #138, opened 2016-06-24
Fix compilation issue under OTP 19

paulcsmith:
There is an issue with ExMachina when used with OTP 19

```
== Compilation error on file test/support/test_factory.ex ==
** (ArgumentError) argument error
    (stdlib) erl_anno.erl:318: :erl_anno.set(:file, 'nofile', -1)
    (stdlib) erl_parse.yrl:1516: anonymous fn/3 in :erl_parse.map_anno/2
    (stdlib) erl_parse.yrl:1631: :erl_parse.modify_anno1/3
    (stdlib) erl_parse.yrl:1517: :erl_parse.map_anno/2
    (stdlib) erl_lint.erl:703: :erl_lint."-set_file/2-lc$^0/1-0-"/2
    (stdlib) erl_lint.erl:448: :erl_lint.exprs_opt/3
    (stdlib) erl_eval.erl:173: :erl_eval.check_command/
```

By removing the `lineL -1` the issue goes away. There is also no warning
when doing this so it can be safely removed.

### DIFF
--- a/lib/ex_machina.ex
+++ b/lib/ex_machina.ex
@@ -173,9 +173,7 @@ defmodule ExMachina do
   end
 
   defmacro __before_compile__(_env) do
-    # We are using line -1 because we don't want warnings coming from
-    # save_record/1 when someone defines there own save_record/1 function.
-    quote line: -1 do
+    quote do
       @doc """
       Raises a helpful error if no factory is defined.
       """


jsteiner:
LGTM!

nathany:
Thanks.

>   # We are using line -1 because we don't want warnings coming from
>   # save_record/1 when someone defines there own save_record/1 function.

I'm just curious what the difference in behaviour will be now (I don't quite understand this aspect of Elixir/Erlang).

Also note:
https://github.com/elixir-lang/elixir/issues/4617#issuecomment-220930717

paulcsmith:
@nathany It appears there is no difference in behavior now because we no longer define factories with `factory` instead it's `factory_#{name_of_factory}`. Also because `save_record` is now deprecated and we are using `ExMachina.Strategy`. So now it's just there as a catch all for those upgrading from old versions that may be calling `factory(:foo)`. So once they upgrade to the new factory definition syntax they *shouldn't* see any warnings. If people do see warnings from this then an issue should be opened, but I don't think they will.

Great question though!

nathany:
Thanks @paulcsmith. Any plans to backport this fix to 0.6.x? (otherwise I may fork it and do that)

Or is master close to a point where 1.0.0-beta.1 could be uploaded to hex as a pre-release?

paulcsmith:
I probably won’t back port it since OTP 19 is super new and 1.0 is about to be released to hex :D https://github.com/thoughtbot/ex_machina/pull/139 <https://github.com/thoughtbot/ex_machina/pull/139>

PR #147, opened 2016-07-05
Only drop autogenerated ids and not domain ids

### LINKED ISSUE #143: Sometimes dropping the primary key isn't desired
Found this problem using `params_for`, etc.

    defp drop_ecto_fields(record = %{__struct__: struct, __meta__: %{__struct__: Ecto.Schema.Metadata}}) do
        ...
        |> Map.drop(struct.__schema__(:primary_key))

In the simple case the primary key is a synthetic id that is auto-generated by the database. In the not so simple case, you have overridden the primary key so that it's not auto-generated by the database. It could be a business key that is meaningful. It could be a one-to-one mapping where the primary key is a foreign key reference to another table.

e.g.
    
    # This has a normal auto-incrementing ID
    schema "plays" do
      has_one :scoring
    end

    # This is a one-to-one modeled in the database by having the PK be a FK to the plays.id column.
    @primary_key {:play_id, :integer, []}
    schema "scoring" do
      field :type, :string
      field :points, :integer

      belongs_to :play, MyApp.Play, define_field: false, references: :id
      belongs_to :team, MyApp.Team, references: :id
    end

In the latter case, the primary key is also a foreign key reference to another table and is a required field, so using `params_for` and friends creates invalid params since it drops the PK.

geofflane:
Fixes #143. Drops only auto-generated ids and not custom IDs that are meaningful in the domain.


test/support/test_factory.ex
```diff
@@ -1,6 +1,12 @@
 defmodule ExMachina.TestFactory do
   use ExMachina.Ecto, repo: ExMachina.TestRepo
 
+  def custom_factory do
+    %ExMachina.Custom{
+      custom_id: 1,
+      name: "Testing"
+    }
+  end
```
paulcsmith: Could you add a newline underneath this?

test/support/test_factory.ex
```diff
@@ -1,6 +1,12 @@
 defmodule ExMachina.TestFactory do
   use ExMachina.Ecto, repo: ExMachina.TestRepo
 
+  def custom_factory do
+    %ExMachina.Custom{
+      custom_id: 1,
```
paulcsmith: maybe `non_autogenerated_id` to make this more clear?

lib/ex_machina/ecto.ex
```diff
@@ -121,9 +121,18 @@ defmodule ExMachina.Ecto do
...
     |> Map.drop(struct.__schema__(:associations))
-    |> Map.drop(struct.__schema__(:primary_key))
+    |> drop_autogenerated_ids(struct)
   end
   defp drop_ecto_fields(record) do
     raise ArgumentError, "#{inspect record} is not an Ecto model. Use `build` instead."
   end
+
+  # Don't want to drop non-autogenrated ids because those are meaningful to the
+  # domain in some way and will be expected to be set explicitly
```
paulcsmith: I think the test explains this well and the comment can be removed

paulcsmith:
This looks awesome! Just a couple small comments and then I'll merge this in

lib/ex_machina/ecto.ex
geofflane: Comment hater! :) I went ahead and removed it...

geofflane:
@paulcsmith Fixed the feedback issues.

paulcsmith:
@geofflane Thank you!

mootpointer:
The above doesn't solve the case where I have an auto-incrementing ID which I have chosen to override. For example if I'm using params_for to generate the parameters that I'm using to update a given record.

Shall I open an issue for this

paulcsmith:
@mootpointer Yes, could you please open an issue with an example of the schema, the factory and the test? I'm not totally sure what you're trying to do so I think that will help us figure it out!

PR #148, opened 2016-07-13
Drop fields that has nil values

Fs02:
This will drop all key that contains nil values which remain untouched previously.
This also fixes error when using ex_machina and [arc_ecto](https://github.com/stavro/arc_ecto) on testing. which is caused by `cast_attachments/3` on arc_ecto doesn't expect the value of the fields to be nil.


lib/ex_machina/ecto.ex
```diff
@@ -133,4 +134,10 @@ defmodule ExMachina.Ecto do
       {name, _type} -> Map.delete(map, name)
     end
   end
+
+  defp drop_nil_values(map) do
+    map
+    |> Enum.filter(fn({_, v}) -> v != nil end)
```
DavidVII: I like this 👍

I feel that using `Enum.reject` would make the intention a bit clearer. What do you think?

Fs02: Yeah, I agree on that. I will update it :smiley:

lib/ex_machina/ecto.ex
```diff
@@ -122,6 +122,7 @@ defmodule ExMachina.Ecto do
     |> Map.delete(:__meta__)
     |> Map.drop(struct.__schema__(:associations))
     |> drop_autogenerated_ids(struct)
+    |> drop_nil_values
```
paulcsmith: Instead of adding this to `drop_ecto_fields` what do you think of adding it as the last pipe in `params_for` and `params_with_assocs`? I think that's more clear than adding it to `drop_ecto_fields`

lib/ex_machina/ecto.ex
```diff
@@ -133,4 +134,10 @@ defmodule ExMachina.Ecto do
       {name, _type} -> Map.delete(map, name)
     end
   end
+
+  defp drop_nil_values(map) do
+    map
+    |> Enum.reject(fn({_, v}) -> v == nil end)
```
paulcsmith: This looks great. Could you change `v` to `value`. We are trying to avoid single letter values so that it's clear to future contributors what's going on

paulcsmith:
@fladson This looks really awesome. Just a couple small comments and then I'll merge this in :D

fladson:
oops, I think you meant @Fs02 😁

paulcsmith:
@fladson Oops! I need to be more careful about using GitHub autocomplete :) Thanks!

Fs02:
I've updated the PR, this is what you meant right?

and I'm wondering whether we should rename `drop_nil_values/1` to `drop_nil_fields` or maybe `drop_empty_fields`?

test/ex_machina/ecto_test.exs
```diff
@@ -50,6 +50,12 @@ defmodule ExMachina.EctoTest do
     end
   end
 
+  test "params_for/2 removes doesn't return nil fields" do
```
paulcsmith: How about "params_for/2 removes fields with nil"?

test/ex_machina/ecto_test.exs
```diff
@@ -77,6 +83,14 @@ defmodule ExMachina.EctoTest do
     }
   end
 
+  test "params_with_assocs/2 doesn't return nil fields" do
```
paulcsmith: And "params_with_assocs/2 remove fields with nil values"

lib/ex_machina/ecto.ex
```diff
@@ -71,6 +71,7 @@ defmodule ExMachina.Ecto do
   def params_for(module, factory_name, attrs \\ %{}) do
     module.build(factory_name, attrs)
     |> drop_ecto_fields
+    |> drop_nil_values
```
paulcsmith: Yeah I think that was a good idea to change this. Maybe `drop_fields_with_nil_valudes`?

paulcsmith:
@Fs02 Yeah that's a good idea to clarify that function name. I left a comment with a suggestion. Once that's addressed I'll go ahead and merge this in!

Fs02:
Updated, thanks!!

paulcsmith:
Looks awesome! Thanks @Fs02

barberj:
New to elixir, but this PR breaks how I was using ex_machina. Curious what the expected approach should be.

I'm setting some valid attrs for testing on a module attribute

```
@valid_attrs FieldGuide.Factory.params_for(:person)
```

And then later on in a test I mutate just what I'm testing and add the required things that the factory didn't provide. 

```
changeset = Person.changeset(%Person{}, %{@valid_attrs | deleted: true, user_id: user.id})
```

Now that user_id is stripped i can't sub it in for my test. Is there a better practice I should be following?

Thanks

paulcsmith:
@barberj I just took a quick looks, but instead of using the update syntax, I think you could use Map.merge

```elixir
changeset = Person.changeset(%Person{}, Map.merge(@valid_attrs, %{deleted: true, user_id: user.id}))
```

Alternatively you could use a function instead of an attribute

```elixir
changeset = Person.changeset(%Person{}, valid_attrs(deleted: true, user_id: user.id))

def valid_attrs(overrides \\ []) do
  FieldGuide.Factory.params_for(:person, Map.new(overrides))
end
```

Final option would be to use `params_for(:person)` directly. Then you can do `params_for(:person, deleted: true, user_id: user.id)`

Does that help?

barberj:
It does, thanks for the help and the awesome elixir libraries

paulcsmith:
Awesome, I'm glad that helped. Thanks for the kind words!

PR #149, opened 2016-07-23
Cast all values before insert

jsteiner:
Casts each field based on the type declared in the schema. Recursively casts built associations.

I added the ability to cast bare maps, but left it separate. After thinking on it some more, I'm not convinced it should be a feature of ExMachina. I couldn't think of a time when you would be building a record with ExMachina, but would want to pass an association in as a bare map. Curious if anybody else has some use cases for that.


lib/ex_machina/ecto_strategy.ex
```diff
@@ -23,4 +25,71 @@ defmodule ExMachina.EctoStrategy do
...
+  end
+
+  defp cast_fields(struct) do
+    struct
+    |> ExMachina.Ecto.drop_ecto_fields
+    |> Map.keys
+    |> cast_fields(struct)
+  end
+
+  defp cast_fields(fields, struct) do
```
paulcsmith: Thoughts on naming this `cast_all_fields` or something like that? I didn't notice the `s` at the end and thought the next function was the same, but with different patterns. Changing the name might help to clarify that a bit

lib/ex_machina/ecto_strategy.ex
```diff
@@ -23,4 +25,71 @@ defmodule ExMachina.EctoStrategy do
...
+      casted_value = cast_field(field, struct)
+      Map.put(struct, field, casted_value)
+    end)
+  end
+
+  defp cast_field(field, %{__struct__: schema} = struct) do
+    field_type = schema.__schema__(:type, field)
+    value = Map.get(struct, field)
+
+    if field_type do
```
paulcsmith: I'm not totally sure I understand this. Would the field ever not have a type? If so, maybe it would help to name this somehow to clarify why?

lib/ex_machina/ecto_strategy.ex
```diff
@@ -23,4 +25,71 @@ defmodule ExMachina.EctoStrategy do
...
+
+  defp cast_assoc(assoc, %{__struct__: schema} = struct) do
+    original_assoc = Map.get(struct, assoc)
+
+    case original_assoc do
+      %{__meta__: %{__struct__: Ecto.Schema.Metadata, state: :built}} ->
+        cast(original_assoc)
+      %{__struct__: _} ->
+        original_assoc
+      records = [_ | _] ->
```
paulcsmith: I'm not sure I understand this. Is it checking if the assoc is a list? If so I think the guard `is_list` might be better here

lib/ex_machina/ecto_strategy.ex
```diff
@@ -23,4 +25,71 @@ defmodule ExMachina.EctoStrategy do
...
+        Enum.map(records, &(cast(&1)))
+      %{} ->
+        assoc_type = schema.__schema__(:association, assoc).related
+        assoc_type |> struct |> Map.merge(original_assoc) |> cast
+      _ ->
```
paulcsmith: Out of curiosity, what would the underscore be. Something like a bare int or a string or something?

paulcsmith:
A few questions/comments. Overall this looks awesome. Can't wait to use this. Will be so handy with dates generated with GoodTimes!

lib/ex_machina/ecto_strategy.ex
jsteiner: Tried `is_list` but it gives me an `illegal pattern` error in the case statement. Rebound the match to `_non_empty_list`, so maybe that helps explain things more?

jsteiner: Actually, I made a couple more changes. Take a look now.

PR #152, opened 2016-08-03
Fix most annoying elixir 1.4 warnings

anronin:
Hi! First of all, thank you for your very nice tool! I just want to fix small but annoying warnings, since i'm using elixir 1.4-dev and phoenix and every time at phoenix starting, i got bunch of warnings. I hope this is not distracting

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -4,7 +4,7 @@ defmodule ExMachina.Mixfile do
   @project_url "https://github.com/thoughtbot/ex_machina"
   @version "1.0.2"
 
-  def project do
+  def project() do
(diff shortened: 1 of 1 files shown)


paulcsmith:
I think this looks good, but I'll need to take install 1.4 dev to double check. Since I don't have that yet I may need a week or two to get to this! Thanks for the PR though. It would definitely be good to make sure we don't emit a bunch of warnings when people use ExMachina under 1.4

paulcsmith:
Thanks for the PR @anronin! This will be in the next release :)

PR #153, opened 2016-08-03
Change `params_for` to return stringify keys

### LINKED ISSUE #146: Should `params_for` return a string keyed map?
I just ran into an issue where my unit tests passed fine but my actual code failed due to a misunderstanding about phoenix and it's params. I wrote the test using ExMachina and the `params_for` which returns atom based keys. When working with the bare params in Phoenix, they come through as string keys. Should we make ex_machina follow this same principle so we can test the same way the framework operates?

drapergeek:
This fixes #146. The general idea is that in Phoenix, params are string keyed, not atom keys. `params_for` returning a map that is not the same as what a controller or changeset normally use, it could cause weird issues between the tests and the actual usage.

Per @paulcsmith we'd like to test this branch out before moving it into the app.

@jsteiner I think you mentioned willing to be a tester.


paulcsmith:
I think it might also be good to convert the topics passed as the second argument to strings. That way when they're merged it is merged correctly

What do you think?

drapergeek:
I'm not sure I understand what you mean by a `topic` in this case.

lIf I pass along any attributes, those are joined in before the stringification happens and so they should be stringified as well unless I'm missing something.

paulcsmith:
I have now corrected my original comment. I meant to type "option" and I have no idea why I typed topic! :P

But yeah I didn't notice that it stringified the keys after the merge so it should be fine as-is!

paulcsmith:
Hey @drapergeek I added a comment on the issue: https://github.com/thoughtbot/ex_machina/issues/146#issuecomment-247629597

TL;DR is that I think we should add a `string_params_for` instead of modifying the behavior of `params_for`. How does that sound?

PR #159, opened 2016-09-02
Replaced usages of the deprecated HashDict module with Map in sequence.ex

cryo28:


lib/ex_machina/sequence.ex
```diff
@@ -1,6 +1,6 @@
 defmodule ExMachina.Sequence do
   def start_link do
-    Agent.start_link(fn -> HashDict.new end, name: __MODULE__)
+    Agent.start_link(fn -> %{} end, name: __MODULE__)
```
pdawczak: How about using [`Map.new/0`](http://elixir-lang.org/docs/stable/elixir/Map.html#new/0) to keep similar level of abstraction?

lib/ex_machina/sequence.ex
```diff
@@ -23,7 +23,7 @@ defmodule ExMachina.Sequence do
      end
   """
   def reset do
-    Agent.update(__MODULE__, fn(_) -> HashDict.new end)
+    Agent.update(__MODULE__, fn(_) -> %{} end)
```
pdawczak: Same as comment above

paulcsmith:
Thank you! I think this is a great change. I like the idea @pdawczak had to use `Map.new`. Once that's switched over I'll go ahead and merge this in!

cryo28:
@paulcsmith Sure 13ecb7c8d317fbed010ef00b31370b4523cd46b1

paulcsmith:
Thank you very much @cryo28!

PR #168, opened 2016-10-10
Add `string_params_for`

pdawczak:
Having seen that the #153 has been closed, I've decided to take that further.

In the #146 issue raised, that `params_for/2` returns a map where the keys are atoms, it makes it difficult to use it in controller tests for Phoenix application. Taking [comment about naming the functions](https://github.com/thoughtbot/ex_machina/issues/146#issuecomment-247629597) into the account, the two new functions have been introduced: `string_params_for/2` and `string_params_with_assocs/2`.

Hope this simple addition makes our lives easier.

What do you think? Is this still valuable?


lib/ex_machina/ecto.ex
```diff
@@ -141,4 +187,10 @@ defmodule ExMachina.Ecto do
     |> Enum.reject(fn({_, value}) -> value == nil end)
     |> Enum.into(%{})
   end
+
+  defp convert_key_atoms_to_strings(struct) do
```
paulcsmith: What do you think of naming this `convert_atom_keys_to_strings`?

lib/ex_machina/ecto.ex
```diff
@@ -94,6 +121,25 @@ defmodule ExMachina.Ecto do
...
+  for Phoenix applications.
+
+  ## Example
+
+      def article_factory do
+        %MyApp.Article{title: "An Awesome Article", author: build(:author)}
+      end
+
+      # Inserts an author and returns %{"title" => "An Awesome Article", "author_id" => 12}
+      params_with_assocs(:article)
```
paulcsmith: I believe this should be `string_params_with_assocs`

lib/ex_machina/ecto.ex
```diff
@@ -94,6 +121,25 @@ defmodule ExMachina.Ecto do
     |> drop_fields_with_nil_values
   end
 
+  @doc """
+  It is similar to `params_with_assocs/2`, but returned map keys are of type
```
paulcsmith: Typically there is a one sentence summary that goes as the first line. How about something like `Similar to `params_with_assocs/2` but converts keys to strings.`

And then the full description below that? Same would go for the other function

lib/ex_machina/ecto.ex
```diff
@@ -75,6 +83,25 @@ defmodule ExMachina.Ecto do
   end
 
   @doc """
+  Works like `params_for/2` but the keys of returned map are strings (instead
+  of atoms). This makes the function useful to be used in controller tests for
+  Phoenix applications where params of this form are expected.
```
paulcsmith: What do you think about "params of this form" -> "where params with string keys" to make it more explicit?

test/ex_machina/ecto_test.exs
```diff
@@ -56,6 +56,13 @@ defmodule ExMachina.EctoTest do
     }
   end
 
+  test "string_params_for/2 produces maps similar to ones built with params_for/2, but the keys are of type string" do
```
paulcsmith: What do you think about changing "are of type string" -> "are strings"

paulcsmith:
@pdawczak Thank you so much for this. I think this would be a great addition, but before merging I'd like to get some input from @jsteiner and @drapergeek 

Nice work!

lib/ex_machina/ecto.ex
pdawczak: Oh... Copy-Paste fail... Thanks!

pdawczak:
Hey @paulcsmith, thanks for your suggestions - I believe I've addressed all of them.

I'm truly glad I can help pushing this great library further! 

In the [related issue](https://github.com/thoughtbot/ex_machina/issues/146) [I've mentioned](https://github.com/thoughtbot/ex_machina/issues/146#issuecomment-247631524) I was thinking about naming those functions, where `params_for` would return _stringified_ version of a map, while there could be `attrs_for` introduced to build maps with atoms.

Unfortunately, this is not backward compatible so if you consider proposed names reflect better the intention of functions, maybe that would b good idea to keep this in mind and set for roadmap for `ExMachina` _v2_?

Thanks!

lib/ex_machina/ecto.ex
```diff
@@ -141,4 +190,10 @@ defmodule ExMachina.Ecto do
     |> Enum.reject(fn({_, value}) -> value == nil end)
     |> Enum.into(%{})
   end
+
+  defp convert_atom_keys_to_strings(struct) do
+    for {key, value} <- struct,
```
jsteiner: @pdawczak I think this needs to be expanded to work for nested maps. Can you also add a test case for that?

Also, minor, but would you mind switching this to use `Enum.reduce` (or something in that family).

paulcsmith:
Thanks @pdawczak!

igorbelo:
any expected date for the release on `hex.pm`?

igorbelo:
is there an estimate to release this in hex.pm?

Issue #169, opened 2016-10-11
has many associations not loaded with params_with_assocs


jbhatab:
Here's my factory

```
def registering_company_factory do
  %Company{
    name: sequence(:name, &"Company #{&1}"),
    email: sequence(:email, &"company-#{&1}@example.com"),
    category: Company.categories |> Enum.random,
    users: [
      %{
        role: Enum.random(User.roles),
      }
    ]
  }
end
```

And I'm calling params_with_assocs(:registering_company) but it only spits out,

%{
  active: nil
  address: nil
  category: "Cultivator"
  description: nil
  email: "company-0@example.com"
  inserted_at: nil
  latitude: nil
  longitude: nil
  name: "Company 0"
  official: nil
  permalinks: [
  ]
  phone: nil
  updated_at: nil
}

Is there a way to load associations with the params call?

hisapy:
Hey guys,

I'm having this problem with `has_many` associations. For example, I have something like the following 

```elixir
defmodule Webapp.SomeStuff do
  schema "some_stuffs" do
    has_many :other_stuffs, Webapp.OtherStuff
  end
end

# and in the news schema
defmodule Webapp.OtherStuff do
  schema "other_stuffs" do
    belongs_to :some_stuff, Webapp.SomeStuff
    # other fields
  end
end

# the factories look look like
def some_stuff_factory do
  %Webapp.SomeStuff{
    news: [build(:other_stuff)]
  }
end

def other_stuff_factory do 
  %Webapp.OtherStuff{
    # fields here
  }
end
```

But when I call `params_with_assoc(:some_stuff)` or `string_params_with_assoc(:some_stuff)` I just get`other_stuffs: [%{}]`. 

`insert` and `build` seem to work just fine.

I'm using **ExMachina v2.1.0**

Issue #93, opened 2016-02-03
fields_for seems to strip has_many through associations


paulcsmith:
``` elixir
fields_for(:post, comments: []) # Results in no goals key
fields_for(:post) |> Map.put(:comments, []) # Sets the results key correctly
```

paulcsmith:
This is caused by https://github.com/thoughtbot/ex_machina/blob/master/lib/ex_machina/ecto.ex#L75. Maybe it makes sense to leave associations and not remove them?

wende:
Not only it'd be cool but it's also a little bit misleading that when you do `params_for(:model)` and you don't have an assoc field that you put there

paulcsmith:
@wende there is a params_with_assocs in ExMachina 1.0 that also includes belongs_to. That might help!

wende:
Yeah. But unfortunately I need `has_many`
I'm inserting an Order with items in it. 
And doing `params_with_assoc(:order)` I expected to have items there. But because they're `has_many` association, they're not there.
And I can't think of a solution to get this without breaking ExMachina's philosophy

paulcsmith:
Ah yes we strip has_many from params in that situation. In this case I would do something like 

params_for(:order) |> Map.put(:order_items, the_items)

Not ideal but might be good for now

PR #174, opened 2016-10-21
Don't delete `has_many`s from param functions

jsteiner:
If you have included a has_many in your factory or passed it into a
param function, then we return those values. We also recursively apply
the same cleanup that we do to the top level struct.

Closes https://github.com/thoughtbot/ex_machina/issues/93
Closes https://github.com/thoughtbot/ex_machina/issues/169


lib/ex_machina/ecto.ex
```diff
@@ -90,9 +91,56 @@ defmodule ExMachina.Ecto do
...
+  end
+
+  defp strip_assoc(original_assoc, record, association_name) do
+    case original_assoc do
+      %{__meta__: %{__struct__: Ecto.Schema.Metadata, state: :built}} ->
+        assoc = recursively_strip(original_assoc)
+        Map.put(record, association_name, assoc)
+      %{__struct__: Ecto.Association.NotLoaded} ->
+        Map.delete(record, association_name)
+      _list ->
```
pdawczak: Have you thought about adding empty line before every pattern to be matched? I't might make it a little bit easier to read, especially, when there are multiple patterns (like in this case).

test/ex_machina/ecto_test.exs
```diff
@@ -44,18 +45,38 @@ defmodule ExMachina.EctoTest do
...
+      articles: [],
     }
   end
 
+  test "params_for/2 recursively strips has_many values" do
+    article = TestFactory.build(:article)
+
+    user_params = TestFactory.params_for(:user, articles: [article])
+
+    assert user_params[:articles] == [%{
```
pdawczak: I appreciate that in these assertions you're trying to express this _nested capabilities_ of what's produced by the factories (in the end, this is what's added in this PR), but would you consider describing _the whole structure_? Something like:

```elixir
assert user_params == %{
  name: "John Doe",
  admin: false,
  articles: [
    %{
      title: "My Awesome Article"
    }
  ],
}
```

could serve documentation purpose in longer term?

Just a thought... 💭

test/ex_machina/ecto_test.exs
```diff
@@ -44,18 +45,38 @@ defmodule ExMachina.EctoTest do
...
+    article = TestFactory.build(:article)
+
+    user_params = TestFactory.params_for(:user, best_article: article)
+
+    assert user_params[:best_article] == %{title: article.title}
+  end
+
+  test "params_for/2 works with maps as has_many values" do
+    article = %{title: "Foobar"}
+    user_params = TestFactory.params_for(:user, articles: [article])
```
pdawczak: In previous tests there is empty line to keep building possible values of `article` before using it in `:user` factory; is it typo here not including it, or in purpose?

test/ex_machina/ecto_test.exs
```diff
@@ -74,20 +95,19 @@ defmodule ExMachina.EctoTest do
     }
   end
 
-  test "params_with_assocs/2 doesn't try to save has_many fields" do
-    assert has_association_in_schema?(ExMachina.User, :articles)
+  test "params_with_assocs/2 keeps has_many associations" do
+    article = TestFactory.build(:article)
+    user_params = TestFactory.params_with_assocs(:user, articles: [article])
```
pdawczak: [Same here](https://github.com/thoughtbot/ex_machina/pull/174/files?diff=split#r84583517)

test/ex_machina/ecto_test.exs
paulcsmith: Good question! The line afterward was used to separate the setup from the exercise and verification phases of the tests https://robots.thoughtbot.com/four-phase-test

test/ex_machina/ecto_test.exs
paulcsmith: I think that's a good thought. Personally, I like just testing the associations because it makes it clear what I should focus on

test/ex_machina/ecto_test.exs
```diff
@@ -44,18 +45,38 @@ defmodule ExMachina.EctoTest do
...
+
+  test "params_for/2 recursively strips has_one values" do
+    article = TestFactory.build(:article)
+
+    user_params = TestFactory.params_for(:user, best_article: article)
+
+    assert user_params[:best_article] == %{title: article.title}
+  end
+
+  test "params_for/2 works with maps as has_many values" do
```
paulcsmith: What do you think about `params_for/2 works with maps that contain has_many associations`?

test/ex_machina/ecto_test.exs
```diff
@@ -44,18 +45,38 @@ defmodule ExMachina.EctoTest do
...
+    article = TestFactory.build(:article)
+
+    user_params = TestFactory.params_for(:user, articles: [article])
+
+    assert user_params[:articles] == [%{
+      title: article.title,
+    }]
+  end
+
+  test "params_for/2 recursively strips has_one values" do
```
paulcsmith: `has_one associations`?

paulcsmith:
A few small comments. LGTM. And thanks for your thoughts @pdawczak!

test/ex_machina/ecto_test.exs
pdawczak: Thank you for your response @paulcsmith!

Yes, _four-phase_ is great patter! In my original comment I was trying to refer to it (though now, when I re-read it, it's not clear - sorry for that).

I was wondering, what's the difference between those two cases, when the [`article` is _built_](https://github.com/thoughtbot/ex_machina/blob/cbc525ec863e3cc05b6b79050e3d2b76ac1255ee/test/ex_machina/ecto_test.exs#L66), and when the [`article` map is prepared _manually_](https://github.com/thoughtbot/ex_machina/blob/cbc525ec863e3cc05b6b79050e3d2b76ac1255ee/test/ex_machina/ecto_test.exs#L74); I have the feeling, that _semantically_ they are exactly the same and as such I'd thought of them to stay in the realm of _setup_ phase (thus, I was asking about the empty line indication).

paulcsmith: Oh you are right. It seems like there should be a newline in between there. I didn't check the code closely enough. Good comment!

test/ex_machina/ecto_test.exs
pdawczak: Yes, I see your point, thanks 👍 !

Just one little thing - the test description mentions _strips has_many values_ (which sounds like _deleting_ them), while assertion here seems to prove opposite. Unless it's just me - would rewording the description to _preserves has_many values/associations_ make it clearer?

pdawczak: Sorry, I think I've misunderstood the description!

The `has_many` association of `User`, the `:articles` is expected to be built, but the _recursive associations_ (eg. `:editor` of `Article`) should not - is this the purpose of this test? In such a case, would the test description `"builds one level of has_many association"` be more helpful? And, as a proof, maybe test like:

```elixir
article = TestFactory.build(:article, editor: build(:user)) # (1). I'm setting editor here explicitly...

user_params = TestFactory.params_for(:user, articles: [article])

assert user_params[:articles] == [%{
  title: article.title,
  editor: nil # (2). ... and proof it's been striped as the second level association
}]
```

What do you think? Would this reveal the intention of the functionality?

KamilLelonek:
Hey, this PR has over 2 months now. When can we expect it to be merged guys?

vraravam:
This might also resolve the [bug](https://github.com/thoughtbot/ex_machina/issues/190) that I just raised. Please merge asap!

jsteiner:
@KamilLelonek @vraravam all set. Sorry about the delay!

Issue #175, opened 2016-10-29
Need a way to customize Ecto structs beyond Map merging


jeffdeville:
I hit a snag when building a multi-tenant solution.  I need to be able to set the prefix when I insert an object.  Something like:

`insert(:user, "tenant_test")`

Problem is that Ecto stores this info in the Ecto.Schema.Metadata.source tuple as {prefix, table_name}

Tuples can't be merged like structs, so Ecto provides: 

`Ecto.put_meta(model, prefix: prefix)`

So the bottom line is that I need to be able to make this call while invoking the insert command.

paulcsmith:
I would take a look at `https://github.com/thoughtbot/ex_machina/blob/master/lib/ex_machina/strategy.ex` and make a custom strategy. You could do something like

`insert_with_tenant(:user, prefix: "tenant_test")` and in your strategy you could grab the prefix from the attrs. Strategies let you do almost anything you want.

LMK if this helps or if you need help figuring out how to use a strategy for this

PR #176, opened 2016-10-29
Allow passing a func to customize structs

jeffdeville:
The 'why' for this is in #175, but in a nutshell, the problem is that merging maps doesn't work in all cases. For example:
- tuples
- nested maps (it doesn't deep-merge)

Tuples are important in the Ecto world, because Ecto.Schema.Metadata.source is a tuple of (prefix, tablename)

More info in the issue:
Resolves: #175

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,26 +117,37 @@ Using factories ([check out the docs](http://hexdocs.pm/ex_machina/ExMachina.htm
 
 ```elixir
 # `attrs` are automatically merged in for all build/insert functions. 
-# Note, a __func__ can be set in the attrs Map. This func will be 
-# executed after the object is created and the rest of the map has been
-# merged. Useful only if you need to alter the generated struct in a 
-# manner that won't work with struct! or Map.merge. 
-# eg: Suppose you need to set the prefix on an object before inserting it:
-insert(:user, %{__func__: fn(user) -> Ecto.put_meta(user, prefix: "prefix") end})
 
 
 # `build*` returns an unsaved comment.
 # Associated records defined on the factory are built.
 attrs = %{body: "A comment!"} # attrs is optional. Also accepts a keyword list.
+func = fn(comment) -> Map.put(comment, :text, "different text") end # func is optional
 build(:comment, attrs)
+build(:comment, func)
+build(:comment, attrs, func)
+
 build_pair(:comment, attrs)
+build_pair(:comment, func)
+build_pair(:comment, attrs, func)
+
 build_list(3, :comment, attrs)
+build_list(3, :comment, func)
+build_list(3, :comment, attrs, func)
 
 # `insert*` returns an inserted comment. Only works with ExMachina.Ecto
 # Associated records defined on the factory are inserted as well.
 insert(:comment, attrs)
+insert(:comment, func)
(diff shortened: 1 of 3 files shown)


jeffdeville:
In case it helps, here's how I'm using this:

```elixir
# in my factory.ex
def test_tenant, do: fn(factory_obj) -> Ecto.put_meta(factory_obj, prefix: "tenant_test") end
def test2_tenant, do: fn(factory_obj) -> Ecto.put_meta(factory_obj, prefix: "tenant_test2") end

# usage

insert(:user, test_tenant)
```

Feels like 'traits' from FactoryGirl

paulcsmith:
I believe that a better option would be to use a new strategy as I mentioned in https://github.com/thoughtbot/ex_machina/issues/175

Could you give that a try? In the meantime I'm going to close this, but I may reopen if a custom strategy doesn't work out

jeffdeville:
Thanks for the feedback @paulcsmith.  That was my first attempt, but perhaps I misunderstood something. To my knowledge, it's not possible to build a struct and pass it in values that don't exist.  (eg: %APP.Model{prefix: "prefix"} will result in a KeyError)

And the strategy method sigs don't accept an additional parameter for options, so I don't see how to make use of strategies in this case.

paulcsmith:
@jeffdeville Ah you are right that it will result in an error since it can't merge keys that don't exist on the struct. However, one option would be to set the prefix as an option.

```elixir
use MyApp.TenantInsertStrategy, prefix: "test_tenant"
```

So by default it always inserts into the same schema. Then for the few tests where you need to explicitly test a different schema, I would use add a function to your factory that you can use so you can do

```elixir
build(:user) |> set_prefix("different_tenant") |> insert
```

This might be a good solution since it sets a default schema, and makes it clear when a test needs a record put in a different schema. I hope that helps/works. LMK if you run into anything :)

jeffdeville:
Hi @paulcsmith, I can appreciate that this is a large PR, and you're understandably loathe to give up the elegant simplicity of a pure map-merge solution. When I first saw the ex_machina src, I was certainly pleased with how approachable and functional it was. 

Lamentably, I've run into a few limitations:

1. Tuples don't merge (this is what bit me)
2. Map merge is only 1 level deep, so I don't think it will produce the expected behavior for nested objects

You're right in saying that I could create 1 tenant, and go vanilla on the other.  I could also create different factories for each tenant, so that I can pass in a different tweak to the custom strategy I also have to write. But to be frank, I can't think of how that solution is as nice as the one I've implemented here.

```
defmodule MyApp.Factory do

  def tenant(name \\ "test"), do: fn(factory_obj) -> Ecto.put_meta(factory_obj, prefix: name) end
  
  def user_factory do
    %MyApp.User{
        name: "Joe"
    }
  end
end

insert(:user, tenant)
insert(:user, tenant("test2"))
```

All that said, here's one other piece of info that might help us arrive at a 3rd alternative. josevalim told me that Ecto 2.1 will allow the user to pass in the prefix into the options of each command. I don't know how this would help us, given that the strategy takes the same parameters as build. It seems like we'd be in the same boat, but perhaps you can see a solution I do not.

PR #179, opened 2016-11-11
Add specs to resolve dialyzer warnings

bernardd:
Applications which include ex_machina and run dialyzer receive warnings
about functions that have no possibility of being called without raising
an exception. Adding these specs fixes those warnings.


paulcsmith:
Thanks so much @bernardd!

PR #188, opened 2017-01-05
Fix a couple of new Elixir 1.4 warnings

lukerollans:
In Elixir 1.4, local arity 0 functions which are called without parentheses raises a warning. Just a quick PR to satisfy these warnings by explicitly using parentheses where applicable.


lukerollans:
PS: On this topic, it seems that there's a lot of warning fixes in `master` that haven't been released yet. Any ETA on this? EG: Swapping out `Map` for `HashDict`

paulcsmith:
Thanks for the PR @lukerollans! Yep there are a few changes in master we need to get done. No ETA, but it should be soon! Just need to make sure everything works as expected in a few of our projects before releasing

PR #189, opened 2017-01-06
docs application dependency update for elixir 1.4

pauloancheta:
Mix v1.4 now automatically infers the applications list
as long as you leave `:applications` key empty


jsteiner:
Thanks!

PR #192, opened 2017-01-26
Drop unbuilt has many throughs in params functions

jsteiner:
Drop unbuilt has many throughs in params functions

Prior to this commit, a has_many through relationship on the model was not dropped from the model. We now delete unbuilt has-many throughs.

Since we already test expected output from various params functions, simply adding the has_many through relation to the model was enough to reproduce errors in the tests.


paulcsmith:
👍

Issue #191, opened 2017-01-14
protocol Enumerable not implemented for nil


vraravam:
Using the latest master of this repo (d2d901e12), my specs fail with the following error:
```elixir
	** (Protocol.UndefinedError) protocol Enumerable not implemented for nil
    (elixir) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir) lib/enum.ex:116: Enumerable.reduce/3
    (elixir) lib/enum.ex:1636: Enum.reduce/3
    (elixir) lib/enum.ex:1188: Enum.map/2
    (ex_machina) lib/ex_machina/ecto_strategy.ex:74: anonymous fn/2 in ExMachina.EctoStrategy.cast_all_assocs/1
    (elixir) lib/enum.ex:1623: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ex_machina) lib/ex_machina/ecto_strategy.ex:17: ExMachina.EctoStrategy.handle_insert/2
```
Please fix asap.

jsteiner:
Can you provide more info on the failing spec, such as the spec and any relevant factories.

vraravam:
@jsteiner - The whole codebase is open source. The specific factory is [here](https://github.com/TWChennai/RecruitX-Backend/blob/master/spec/support/factory.ex).

jsteiner:
I've added a failing test to reproduce this here: https://github.com/thoughtbot/ex_machina/compare/js-nil

PR #193, opened 2017-01-27
Allow setting associations to nil

jsteiner:
Previously the catchall was falling through to the list case and trying
to enumerate `nil`. This commit adds a guard to the list condition to
ensure that doesn't happen again, then adds a new condition to handle
`nil`.

I _think_ this should be handling all cases now, so it also removes the
catchall. This way, if it fails in a way we didn't expect, it should
give an obvious failure that there was no matching case.

Fixes https://github.com/thoughtbot/ex_machina/issues/191

### DIFF
--- a/lib/ex_machina/ecto_strategy.ex
+++ b/lib/ex_machina/ecto_strategy.ex
@@ -82,13 +82,19 @@ defmodule ExMachina.EctoStrategy do
     case original_assoc do
       %{__meta__: %{__struct__: Ecto.Schema.Metadata, state: :built}} ->
         cast(original_assoc)
+
       %{__struct__: _} ->
         original_assoc
+
       %{} ->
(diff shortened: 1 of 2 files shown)


lib/ex_machina/ecto_strategy.ex
```diff
@@ -82,13 +82,19 @@ defmodule ExMachina.EctoStrategy do
...
       %{} ->
         assoc_type = schema.__schema__(:association, assoc).related
         assoc_type |> struct |> Map.merge(original_assoc) |> cast
-      _list ->
+
+      list when is_list(list)->
         Enum.map(original_assoc, &(cast(&1)))
+
+      nil ->
+        nil
```
paulcsmith: Nice catch!

PR #196, opened 2017-02-04
Clarify where to import factory module in Phoenix

arsduo:
Being new to Phoenix I spent some time this morning trying to figure out why I got the error `undefined function insert/1` when setting up ExMachina. Turns out I'd put the `import MyApp.Factory` into the body of MyApp.ConnCase instead of in the `using` block. 

This PR adds a note to the readme in the hope of saving other developers the time it took to figure that out.

Thanks for the great library!

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,8 @@ string_params_with_assocs(:comment, attrs)
 # Example of use in Phoenix with a factory that uses ExMachina.Ecto
 defmodule MyApp.MyModuleTest do
   use MyApp.ConnCase
-  # You can also import this in your MyApp.ConnCase if using Phoenix
+  # You can also import this in your MyApp.ConnCase (inside the using block)
+  # if using Phoenix
   import MyApp.Factory
 
   test "shows comments for an article" do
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -152,7 +152,8 @@ string_params_with_assocs(:comment, attrs)
 # Example of use in Phoenix with a factory that uses ExMachina.Ecto
 defmodule MyApp.MyModuleTest do
   use MyApp.ConnCase
-  # You can also import this in your MyApp.ConnCase if using Phoenix
+  # You can also import this in your MyApp.ConnCase (inside the using block)
+  # if using Phoenix
```
paulcsmith: Maybe we could shorten this to `If using Phoenix, import this inside the using block in MyApp.ConnCase`. I think that should fit on one line

paulcsmith:
@arsduo Great idea! I had just one comment. Other than that this looks great :D

arsduo:
Thanks for the feedback! I've updated the wording as you suggested.

paulcsmith:
Wonderful change. Thanks for submitting it!

PR #201, opened 2017-03-03
Bump to 2.0.0

jsteiner:


paulcsmith:
We are getting so close :D It might be worth doing a find and replace on `1.0`. There are a few places like the installation instructions in the README that still have the old version.

PR #203, opened 2017-03-06
Fix link to latest published docs

ngrash:
I clicked on the link to the *docs for the latest published version of ExMachina* and hexdocs.pm gave me a 404. Turns out `README.html` should be `readme.html`. This PR fixes the link.

Let me know if I should open an issue. I did not open an issue in the first place, because I felt this is a really minor fix.


jsteiner:
Thanks!

PR #204, opened 2017-03-10
Strip embeds in params_for

leonid-shevtsov:
I had an issue that `params_for` didn't work for my model that had embed fields in it; you have to strip embed structs down to plain maps for them to be allowed inside a changeset.

I removed the pattern matching dealing with `Ecto.Schema.Metadata` because `embedded_schema`s don't have metadata; it should always be safe as all traversed objects have Ecto schemas by the context of the `recursively_strip` method.

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -152,18 +152,12 @@ defmodule ExMachina.Ecto do
   end
 
 
-  defp recursively_strip(record = %{__meta__: %{__struct__: Ecto.Schema.Metadata}}) do
+  defp recursively_strip(record = %{__struct__: _}) do
     record
     |> set_persisted_belongs_to_ids
     |> handle_assocs
     |> handle_embeds
     |> drop_ecto_fields
-    |> drop_fields_with_nil_values
-  end
-
-  defp recursively_strip(embedded_record = %{__struct__: _}) do
-    embedded_record
-    |> Map.from_struct
     |> drop_fields_with_nil_values
   end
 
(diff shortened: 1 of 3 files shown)


pdawczak:
Would that be beneficial to add tests covering this work also for `params_with_assocs/2`, `string_params_for/2`, `string_params_with_assocs/2`?

pdawczak:
I believe it will address this issue: https://github.com/thoughtbot/ex_machina/issues/210

leonid-shevtsov:
Rebased from latest master to resolve conflicts

lib/ex_machina/ecto.ex
```diff
@@ -191,7 +192,30 @@ defmodule ExMachina.Ecto do
...
+  defp handle_embed(original_embed, record, embed_name) do
+    case original_embed do
+      %{__struct__: _} ->
+        embed = recursively_strip(original_embed)
+        Map.put(record, embed_name, embed)
+
+      list when is_list(list) ->
+        embeds_many = Enum.map(original_embed, &recursively_strip/1)
+        Map.put(record, embed_name, embeds_many)
+
```
paulcsmith: Could you please remove the newline here and on line 208 to keep with existing style?

test/ex_machina/ecto_test.exs
```diff
@@ -105,6 +105,25 @@ defmodule ExMachina.EctoTest do
     assert user_params.articles == [%{title: article.title}]
   end
 
+  test "params_for/2 strips embeds_one" do
```
paulcsmith: Maybe "cleans" instead of "strips" would make more sense. I was a bit confused at first because I thought it was completely removing the embeds

leonid-shevtsov:
@paulcsmith Done!

PR #206, opened 2017-03-13
Ecto: Support bare maps as values for embeds. Support lists of bare maps

leonid-shevtsov:
The first one was only supported for assocs. 

Passing a bare map for an embed first crashed in `cast_value` because it can't be cast trivially, then in `cast_assoc` because the bare map handling code expects an association only.

The second one wasn't supported for either assocs or embeds, because bare maps inside of the list weren't properly casted.

Added tests and fixed.


jsteiner:
@leonid-shevtsov thanks!

PR #209, opened 2017-03-15
Fix broken build due to typos

rbeene:
* recrod -> record
* Kernal -> Kernel


paulcsmith:
Thanks for catching this!

PR #220, opened 2017-05-26
Minor clean up of EctoStrategy

germsvel:
* Renamed `embed_field?` variable to `embedded_field?`
* Extracted branching logic into `cast_assoc/3` for the case when the association is
a list.

### DIFF
--- a/lib/ex_machina/ecto_strategy.ex
+++ b/lib/ex_machina/ecto_strategy.ex
@@ -74,19 +74,17 @@ defmodule ExMachina.EctoStrategy do
     assocs = get_schema_assocs(schema)
 
     Enum.reduce(assocs, struct, fn(assoc, struct) ->
-      casted_value = Map.get(struct, assoc)
-                   |> cast_assoc(assoc, struct)
+      casted_value = struct |> Map.get(assoc) |> cast_assoc(assoc, struct)
 
       Map.put(struct, assoc, casted_value)
     end)
   end
 
-  defp cast_assoc(original_value, assoc, struct) when is_list(original_value) do
-    Enum.map(original_value, &(cast_assoc(&1, assoc, struct)))
-  end
-
-  defp cast_assoc(original_value, assoc, %{__struct__: schema}) do
+  defp cast_assoc(original_value, assoc, %{__struct__: schema} = struct) do
     case original_value do
+      original_value when is_list(original_value) ->
+        Enum.map(original_value, &(cast_assoc(&1, assoc, struct)))
+
       %{__meta__: %{__struct__: Ecto.Schema.Metadata, state: :built}} ->
         cast(original_value)
 
@@ -99,7 +97,7 @@ defmodule ExMachina.EctoStrategy do
       %{} ->
         assoc_reflection = schema.__schema__(:association, assoc) || schema.__schema__(:embed, assoc)
         assoc_type = assoc_reflection.related
-        assoc_type |> struct |> Map.merge(original_value) |> cast
+        assoc_type |> struct() |> Map.merge(original_value) |> cast
 
       nil -> nil
     end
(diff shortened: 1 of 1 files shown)


lib/ex_machina/ecto_strategy.ex
```diff
@@ -74,16 +74,17 @@ defmodule ExMachina.EctoStrategy do
...
-        cast_assoc(original_value, assoc, struct)
-      end
+      casted_value = Map.get(struct, assoc)
+                   |> cast_assoc(assoc, struct)
+
       Map.put(struct, assoc, casted_value)
     end)
   end
 
+  defp cast_assoc(original_value, assoc, struct) when is_list(original_value) do
```
jsteiner: I'd probably move this back into the `case` statement below so the forking is all done in one place. Having one of these as a separate function and the others in a `case` seems sort of arbitrary.

lib/ex_machina/ecto_strategy.ex
```diff
@@ -74,16 +74,17 @@ defmodule ExMachina.EctoStrategy do
...
 
     Enum.reduce(assocs, struct, fn(assoc, struct) ->
-      original_value = Map.get(struct, assoc)
-      casted_value = if is_list(original_value) do
-        Enum.map(original_value, &(cast_assoc(&1, assoc, struct)))
-      else
-        cast_assoc(original_value, assoc, struct)
-      end
+      casted_value = Map.get(struct, assoc)
+                   |> cast_assoc(assoc, struct)
```
jsteiner: Maybe `casted_value = struct |> Map.get(assoc) |> cast_assoc(assoc, struct)`

lib/ex_machina/ecto_strategy.ex
germsvel: Yeah, I was thinking about that too. I guess the difference is that here the `original_value` is a list as opposed to an map. The rest of the branching logic is dealing maps, one way or another, so it fits more nicely. That being said, I can see why we'd want it in a single place. 

Do you think this is a preferable? 

```elixir
case original_value do 
  original_value when is_list(originaLvalue) -> 
    Enum.map(original_value, &(cas_assoc(&1, assoc, struct)))
  
  %{__meta__: %{__struct__: Ecto.Schema.Metadata, state: :built}} ->
    cast(original_value)

  %{__struct__: Ecto.Association.NotLoaded} ->
    original_value

  %{__struct__: _} ->
    cast(original_value)

  %{} ->
    
# ...

```

lib/ex_machina/ecto_strategy.ex
germsvel: Yeah, that's nicer. I'll update to do this.

lib/ex_machina/ecto_strategy.ex
jsteiner: It feels like an improvement to me, although I don't feel super strongly.

germsvel: Yeah, I think you're right. Fixed in fecae08411ba7964bfa1f765352e30747ccba24c

lib/ex_machina/ecto_strategy.ex
```diff
@@ -74,18 +74,17 @@ defmodule ExMachina.EctoStrategy do
...
+      casted_value = struct |> Map.get(assoc) |> cast_assoc(assoc, struct)
+
       Map.put(struct, assoc, casted_value)
     end)
   end
 
-  defp cast_assoc(original_value, assoc, %{__struct__: schema}) do
+  defp cast_assoc(original_value, assoc, %{__struct__: schema} = struct) do
     case original_value do
+      original_value when is_list(original_value) ->
```
jsteiner: Maybe `list when is_list(list)` to avoid shadowing?

jsteiner:
One more small comment, but LGTM

lib/ex_machina/ecto_strategy.ex
germsvel: Hmmm. yeah. It looks a bit weird. Fixed in e03690a6bd357c11dc0b0d5c2df648686f6a96a1

Issue #222, opened 2017-06-01
`string_params_for` does not stringify keys in associations


zwippie:
- elixir 1.4.1
- phoenix 1.3.0-rc.2
- ecto 2.1.4
- ex_machina 2.0.0

Given a Phoenix + Ecto app with the following factory:

```
def album_factory do
  %Album {
    name: sequence(:name, &"Album #{&1}"),
    items: build_list(5, :item)
  }
end

def item_factory do
  %Item {
    name: sequence(:name, &"item-#{&1}")
  }
end
```

The produced output of `string_params_for(:album)` is:

```
iex(13)> string_params_for(:album)
%{"items" => [%{name: "item-49"}, %{name: "item-50"}, %{name: "item-51"},
   %{name: "item-52"}, %{name: "item-53"}], "name" => "Album 48"}
```

The expected output should have the `name` key of every associated `item` as a string, not an atom:

```
%{"items" => [%{"name": "item-49"}, %{"name": "item-50"}, %{"name": "item-51"},
   %{"name": "item-52"}, %{"name": "item-53"}], "name" => "Album 48"}
```

Which is understandable because of the call to `build_list` in `album_factory`, but it's confusing none the less.

zwippie:
I just realized that the usage of stringified keys in associations within Phoenix controller tests (the apparent use case for the `string_params` functions)  is quite weak; ecto does not accept associations as flexible as for example ActiveRecord does.

So this might just be a non-issue, but it's a little confusing none the less.

germsvel:
@zwippie thanks for reporting this. Could you take a look at https://github.com/thoughtbot/ex_machina/pull/224 and see if it solves the issue?

germsvel:
I believe https://github.com/thoughtbot/ex_machina/pull/224 closes this issue. If not, please re-open. Thank you for reporting this!

zwippie:
Seems to work perfect, thanks for fixing this!

Issue #215, opened 2017-04-25
string_params_for should cast embeddings into maps


Sitch:
Lets say you have a `Car` that `embeds_one` `Engine`.

```elixir
def car_factory do
  %Car{
    engine: build(:engine)
  }
end
```
if you call `string_params_for`, the nesting will still have a struct typed `Engine` instead of a plain map. i.e:

```elixir
%{
  "engine" => %Engine{ ... }
}
```
not
```elixir
%{
  "engine" => %{ ... }
}
```

germsvel:
Hi @Sitch, I find that a little counterintuitive since `string_params_for(:car)` doesn't give me `%Car{}` with string keys but `%{}` with string keys. So if that's the behavior at the top level, I would think we would keep that behavior consistent when dealing with nested structures. What do you think?

Sitch:
@germsvel the use case for this post was getting a semi-deep nested json payload from a 3rd party provider.  For me, it was useful to not have to write the same factory in two different ways (one version with typed nested structures and one version that represented the json decoding with string keys).  Additionally, the rational in the docs for `string_params_for` is phoenix controllers, which would not implicitly type cast those nested arguments

germsvel:
@Sitch I'm sorry. I misread your original post. I thought you _wanted_ `string_params_for/2` to give, 

```elixir
%{
  "engine" => %Engine{ ... }
}
```

I have now read it correctly. You're saying that's the _current_ behavior, but that instead it should give you something like this, 

```elixir
 %{
  "engine" => %{ ... }
}
```

I agree. Sorry for the confusion.

Sitch:
@germsvel no worries -- gotta work on my verbal clarity 👓.  Let me know if you want help with the PR

germsvel:
@Sitch thinking about it, I actually think https://github.com/thoughtbot/ex_machina/pull/224 will fix this issue.

germsvel:
I believe #224 closes this. If not, please re-open. Thank you for reporting this!

PR #224, opened 2017-06-09
Handle nested structures in `string_params_for/2`

germsvel:
Add support for `.string_params_for/2` to handle nested associations and
nested embeds: `has_one`, `has_many`, `embeds_one`, and `embeds_many`.

I believe this addresses https://github.com/thoughtbot/ex_machina/issues/222, and https://github.com/thoughtbot/ex_machina/issues/215

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -279,9 +279,13 @@ defmodule ExMachina.Ecto do
     |> Enum.into(%{})
   end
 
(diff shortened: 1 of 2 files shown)


Sitch:
One exception that could be worth mentioning here is for `%Plug.Upload{}` if the intended purpose of `string_params_for` is for controller testing

germsvel:
I don't know if we should handle `%Plug.Upload{}` here explicitly since this is ecto-specific but not necessarily related to phoenix.

Issue #221, opened 2017-06-01
Naively inserting a record with children doesn't work


bencates:
First, thanks for the awesome project!

I'm working on an app that has two Ecto schemas, `Order` and `Product`. Orders have many products, and products belong to orders. I have the following factories:

```elixir
def order_factory() do
  %MyApp.Order{
    # snip
  }
end

def product_factory() do
  %MyApp.Product{
    # snip
    order: build(:order),
  }
end
```

In my tests I attempt to insert an order and a list of products like so:

```elixir
order = insert(:order, products: [build(:product)])
```

and I get the somewhat cryptic error "** (ArgumentError) cannot change belongs_to association `order` because there is already a change setting its foreign key `order_id` to `3305`".

This works, though:

```elixir
order = insert(:order, products: [build(:product, order: nil)])
```

It's not a terribly difficult workaround, but it adds a significant amount of irrelevant noise to my tests, especially on tests which require deeply nested objects.

tomekowal:
Thank you @bencates for posting this workaround! My current solution was to insert order first and then product which was even more noisy. It would be great if inserting list in `has_many` association worked out of the box. If I have some time this week, I'll try to familiarize with `ex_machina` internals and check how much time it would get me to fix the issue.

tomekowal:
Hi! I don't have full solution yet, but I have found a way to reproduce the problem.
`ex_machina` has a test for `has_many` association here:
https://github.com/thoughtbot/ex_machina/blob/master/test/ex_machina/ecto_strategy_test.exs#L55
but it passes only by accident.

In model definitions:

    schema "users" do
      ...
      has_many :articles, ExMachina.Article,
      ...
    end

and:

    schema "articles" do
      ...
      belongs_to :author, ExMachina.User
      ...
    end

`has_many` association will search `user_id` in `Article`, but `Article` defines `author_id`, so in the test when we do:

    built_article = TestFactory.build(:article, visits: 10)
    model = TestFactory.insert(:user, articles: [built_article])

there are actually two users created instead of one. If we fix the schema by referencing correct foreign key:

    schema "users" do
      ...
      has_many :articles, ExMachina.Article, foreign_key: :author_id,
      ...
    end

we will get the same error as @bencates posted above. I've made clone with branch that demonstrates the problem (two tests don't pass after fixing the schema), but I don't think it makes sense to turn it into a PR before I have a fix.
https://github.com/tomekowal/ex_machina/tree/fix-user-has-many-posts-association

[...]

PR #225, opened 2017-06-14
Fix user has many posts association

tomekowal:
This is my first stab at solving #221 

In case where user has many articles and `article_factory` defines `build(:user)`, it should be still possible to call it like this:

    Factory.insert(:user, name: "parent user", articles: [build(:article)])

In that case `ex_machina` should ignore the `build(:user)` part of `article_factory`, because it is already set by its parent `user` in the tree of structs that we want to isnert.

This PR extends `EctoStrategy.cast` function to take additional argument `field_set_by_parent`.
This argument is set to foreign key of `has_many` relation. (I am pretty sure we will need another case for `has_one` relation, but we can go one step at a time).
Foreign key is the only thing that connects relation "User has_many Articles" with relation "Article belongs_to User" that we want to skip.

Please let me know if you like this solution and if you think I can make it better in any way.
Thanks in advance for the time spent on review!

### DIFF
--- a/lib/ex_machina/ecto_strategy.ex
+++ b/lib/ex_machina/ecto_strategy.ex
@@ -31,8 +31,8 @@ defmodule ExMachina.EctoStrategy do
 
     record
     |> cast_all_fields
-    |> cast_all_assocs(field_set_by_parent)
     |> skip_already_set_assoc(parent_assoc)
+    |> cast_all_assocs(nil)
   end
 
   defp assoc_set_by_parent(%{__struct__: schema}, field_set_by_parent) do
--- a/lib/ex_machina/ecto_strategy.ex
+++ b/lib/ex_machina/ecto_strategy.ex
@@ -14,7 +14,7 @@ defmodule ExMachina.EctoStrategy do
 
   def handle_insert(%{__meta__: %{__struct__: Ecto.Schema.Metadata}} = record, %{repo: repo}) do
     record
-    |> cast
+    |> cast(nil)
     |> repo.insert!
   end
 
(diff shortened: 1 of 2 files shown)


febeling:
Any thoughts on this PR?

tlvenn:
@germsvel any chance to review this PR ?

tlvenn:
@germsvel is your work in 3776a51b3ad72fafc9d840471c245c8ff9b7ea13 supersede this PR somewhat ? It would seem on the surface that by leveraging Ecto metadata, ExMachina can deal properly with any associations and figure out if it needs to persist them or not ?

Can you please clarify the situation ? Thanks in advance.

germsvel:
Hey @tlvenn, my work on that commit does not supersede this PR. It just simply touches some of the same files. 

@tomekowal thanks so much for submitting this PR, for your comments in the associated issue, and for finding that bug in the test setup! I have taken a look at this PR, and though I see it works, this issue has caused me to think about a broader question of how we want to handle associations. Maybe the approach we have right now works fine and then a fix like this one would be okay, but I'm trying to see if this issue hints at a need for a different approach.

In either case, I hope to open an issue with some possibilities when I have gathered my thoughts a little bit more, and maybe people can chime in there. Until then, I hope the workarounds are working okay. And thanks for your patience!

tomekowal:
Thanks for your reply! I am happy to discuss different approaches for handling associations. If you find better solution for those problems I may also help with coding tasks.

axelson:
Any chance this can be reviewed? Is there anything that I can do to move this forward?

germsvel:
@tomekowal I haven't had the time to dig into thinking of associations in a different way, like I had suggested. But this issue seems to be something that trips up a lot of people, so I think we should get this in. 

I know it's been a while so feel free to say "no way", but would you be interested in resolving conflicts and trying to get this change merged? If you're not up for it, I can reimplementing this in another PR. I'd just like to give you the opportunity to get this commit in since you worked on it, and I appreciate the time you spent on it!

tomekowal:
@germsvel I've looked into it and I am not sure that my initial solution was 100% correct on two levels:

a) since the relations are not bidirectional I've nilified foreign key of `has` relation. Example:
`%User{articles: [%Article{author: %User{}}]}` would set `Article.author_id` to `nil` which has an intended side effect that `author` is ignored. However, I am not sure if the "inner" user is also persisted. In such a case, the factory would insert two users and the intended result is to insert only parent user (the outer one) and ignore the inner one because it is there only as a default because of `author: build(:user)` in `article_factory`. It was very "hacky"

b) Even if I am wrong in point a) and it works as intended, it is still not explicit. The inner user is discarded which might be as confusing as the error.

I believe the correct way to solve the problem is to explicitly say not to build the inner user record.

```
  test "insert/1 casts has_many associations" do
    built_article = TestFactory.build(:article, visits: 10, author: nil)
    model = TestFactory.insert(:user, articles: [built_article])

    assert List.first(model.articles).visits == Decimal.new(10)
  end
```

I thought about different way to solve that problem, but nothing comes to my mind. I'll make another PR which only fixes the tests setup.

germsvel:
@tomekowal thanks so much for taking the time to look through that again!

Regarding (a), I think you're right in that depending on the implementation, we may persist two users, and orphan one of them. And I agree that it seems a bit hacky to do that. 

---

> Even if I am wrong in point a) and it works as intended, it is still not explicit. The inner user is discarded which might be as confusing as the error.

I totally agree. I prefer not having to delete data that a user of ExMachina may have desired to insert (the inner author). 

---

> I believe the correct way to solve the problem is to explicitly say not to build the inner user record.

Unfortunately, I think that's the case for now. 

I also think another okay solution is to just `insert` the first record instead of building it. I know that would have the side effect of persisting an extra author, but at least the factories are behaving as expected:

```elixir
    article = insert(:article, visits: 10)
    user = insert(:user, articles: [article])
```

Thanks again for looking through this! I'll close this for now, but let me know if you want to reopen this.

PR #227, opened 2017-06-28
So that we can use list sequences, fixes #223

### LINKED ISSUE #223: Feature Idea: Enums
It would be great if we could do something like:
```elixir
def role_factory do
  %Role{
    name: enum(["owner", "admin", "guest"])
  }
end
```
which would behave like:
```elixir
build(:role)
# %Role{name: "owner"}
build(:role)
# %Role{name: "admin"}
```

suhrawardi:
Now we can use list sequences, as proposed in #223

```
  def user_factory do
    %MyApp.User{
      name: "Jane Smith",
      email: sequence(:email, &"email-#{&1}@example.com"),
      role: sequence(:role, ["admin", "user", "other"]),
    }
  end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ defmodule MyApp.Factory do
     %MyApp.User{
       name: "Jane Smith",
       email: sequence(:email, &"email-#{&1}@example.com"),
+      role: sequence(:role, ["admin", "user", "other"]),
     }
   end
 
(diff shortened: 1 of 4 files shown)


suhrawardi:
@paulcsmith, sure, added it!
Must have overlooked that one... :-)

germsvel:
Thanks for this PR @suhrawardi!

PR #230, opened 2017-07-18
Fixes issue with passing maps for embeds to params_for/2

kbredemeier:
I just tried out using the master branch for embeds in `params_for/2` and encountered unexpected behavior. In my current project I already have calls like `prams_for(:product, weight: %{value: 10, unit: "kg"})` and this fails in `recursively_strip/1`.

This PR fixes the issue.


jsteiner:
Thanks!

PR #232, opened 2017-07-18
handle the number 0 in build_list and insert_list

### LINKED ISSUE #231: build_list and insert_list should handle the number 0
Currently, both functions return 2 records when asked to generate 0 records.

This happens because the range 1..0 is 2 elements long.

Here is the context where I encountered this problem: 
```elixir
insert_list(Enum.random(0..10), :foo)
```

I have a PR with the fix coming right up!

alxgnon:
fixes #231

### DIFF
--- a/test/ex_machina/ecto_test.exs
+++ b/test/ex_machina/ecto_test.exs
@@ -30,6 +30,10 @@ defmodule ExMachina.EctoTest do
     assert [%User{}, %User{}, %User{}] = TestFactory.insert_list(3, :user, admin: true)
   end
 
+  test "insert_list/3 handles the number 0" do
+    assert [] = TestFactory.insert_list(0, :user)
+  end
+
   test "params_for/2 removes Ecto specific fields" do
     assert TestFactory.params_for(:user) == %{
       name: "John Doe",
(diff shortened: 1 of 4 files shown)


test/ex_machina/ecto_test.exs
```diff
@@ -30,6 +30,10 @@ defmodule ExMachina.EctoTest do
     assert [%User{}, %User{}, %User{}] = TestFactory.insert_list(3, :user, admin: true)
   end
 
+  test "insert_list/3 handles the number 0" do
+    assert [] = TestFactory.insert_list(0, :user, admin: true)
```
jsteiner: `admin: true` is unnecessary. Would you mind dropping it?

test/ex_machina_test.exs
```diff
@@ -99,6 +99,10 @@ defmodule ExMachinaTest do
     assert records == [expected_record, expected_record, expected_record]
   end
 
+  test "build_list/3 handles the number 0" do
+    assert [] = Factory.build_list(0, :user, admin: true)
```
jsteiner: Same here.

alxgnon:
@jsteiner fixed.

jsteiner:
Thanks!

Issue #202, opened 2017-03-03
Add documentation for `insert` functions


jsteiner:
Since these functions are generated dynamically, we don't have any docs for them. Ecto solves this with the `@callback` directive: https://github.com/elixir-ecto/ecto/blob/v2.1.3/lib/ecto/repo.ex#L735

schrockwell:
Additionally, the documentation for the `build` functions should also use the `@callback` spec, since right now the function signature of, for example, `ExMachina.build/3` does not mach the code samples, since it's not the same same as the `build/2` which is injected to a module with `use ExMachina`.

PR #233, opened 2017-07-28
Improving documentation

schrockwell:
Hi folks. I've been diving into ExMachina lately and found the docs a little confusing. Here's my initial attempt to improve them. 

Similar to how the `Ecto.Repo` docs work, I've added callback docs for injected functions like `build/2`, etc so that the function signatures match up with the code examples. I also added `insert/2` et. al. docs for `ExMachina.Ecto` using the same method, since those were missing entirely (see #202).

- [x] Add callback docs for injected factory functions
- [x] Add docs for `ExMachina.Ecto`
- [x] Add missing typespecs


jsteiner:
This is awesome. Thanks @schrockwell!

PR #237, opened 2017-08-09
New version of ecto formats autogenerated id slightly differently.

zachdaniel:
In their new RC (v2.2.0-rc.0), they changed the shape of this.


suhrawardi:
Thx @zachdaniel, would be very happy to have this merged in!

paulcsmith:
@zachdaniel I merged #238 instead, but thank you for making this. You should be able to use master now. Hoping to get a release out this week so you can use from hex instead of GitHub

PR #239, opened 2017-08-21
Add example of dependent factory

nathanl:
This may seem obvious, but it took me a few minutes to figure out and I thought it would be nice to show.


README.md
```diff
@@ -104,6 +104,14 @@ defmodule MyApp.Factory do
     }
   end
 
+  # dependent factory
```
paulcsmith: I think `# Modify existing factories` might be more clear here. What do you think?

README.md
```diff
@@ -104,6 +104,14 @@ defmodule MyApp.Factory do
...
+  def featured_article_factory do
+    %MyApp.Article{
+      article_factory |
+      featured: true,
+    }
```
paulcsmith: I think this syntax is a bit nicer. What do you think? `struct!(article_factory, %{featured: true})`

README.md
nathanl: Maybe `derived factories`? That's something I searched for first.

nathanl:
Rebased on master.

Issue #243, opened 2017-08-30
string_params_for with structs


Sitch:
I am using the money struct and i get enumerable errors when calling `string_params_for`

Error:
```
** (Protocol.UndefinedError) protocol Enumerable not implemented for %Money{amount: 1200000, currency: :USD}
        (elixir) lib/enum.ex:1: Enumerable.impl_for!/1
        (elixir) lib/enum.ex:116: Enumerable.reduce/3
        (elixir) lib/enum.ex:1767: Enum.reduce/3
    (ex_machina) lib/ex_machina/ecto.ex:321: anonymous fn/2 in ExMachina.Ecto.convert_atom_keys_to_strings/1
        (stdlib) lists.erl:1263: :lists.foldl/3
```

This is with master branch

Sitch:
basically we just need to call: `Enumerable.impl_for(map)` before we traverse the map

germsvel:
This should be fixed by https://github.com/thoughtbot/ex_machina/pull/245, so I will close it. Let me know if that's not the case, and we can reopen it.

PR #244, opened 2017-08-31
Bump version to 2.1.0

paulcsmith:


paulcsmith:
A lot of people are eager to see this released so if one of you could review this tomorrow morning that would be awesome! Once it's in I'll publish a new release

CHANGELOG.md
```diff
@@ -7,6 +7,27 @@ complete changelog, see the git history for each version via the version links.
 
 [hex package page]: https://hex.pm/packages/ex_machina
 
+## [2.1.0]
+
+### Breaking changes
+
+- Supports only Ecto 2.2 or higher https://github.com/thoughtbot/ex_machina/commit/8f494f1b72612ac4784f801e0f658ce30b4b3552
```
jsteiner: I believe this means we need to update to `3.0.0`. Also, I believe there's another breaking change, but I can't remember what it was. @germsvel probably knows more.

germsvel: Yes, there was another breaking change that I was hoping to fix but haven't had the time. Someone actually opened an issue about it https://github.com/thoughtbot/ex_machina/issues/243

germsvel: I don't know if only supporting Ecto 2.2 means that we should bump to 3.0., unless it breaks ex_machina's public api. This is from [semver.org](http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api):

> What should I do if I update my own dependencies without changing the public API?

> That would be considered compatible since it does not affect the public API. Software that explicitly depends on the same dependencies as your package should have their own dependency specifications and the author will notice any conflicts.

paulcsmith: @jsteiner I also just remembered that the PR I merged maintains support for Ecto 2.1. So it's actually not a breaking change in this release. So I think 2.1.0 works

PR #247, opened 2017-09-06
Expose EctoStrategy.cast/1

### LINKED ISSUE #258: [RFC] Exposing EctoStrategy.cast/1 to support view insertion
We have many views in our system and would like to utilize the ease of factories when testing them.

In practice, this requires us to create additional insert functions, such as `insert_view/1`, which takes the view backed schema as its sole argument. Within the function, we define all the logic required to persist the view; i.e. population of all the component tables. The one stumbling block though, is casting the schema to get all the ex_machina magic. We've worked around the issue by exposing `EctoStrategy.cast/1` in our fork.

Is this a proper approach, or are we going about this all wrong?

If it is agreed this approach is suitable, we have submitted a PR to address.

erikreedstrom:
This change allows the `cast` function to be invoked outside of the strategy.

Invoking `cast/1` is useful for complex multi or schemaless inserts involving views,
where the struct can be treated the same as other Ecto schemas without having
to recreate the full `Ecto` and `EctoStrategy` modules.

## Example

As an additional factory method:
```elixir
# Location is a view backed schema
def insert_view(%Location{} = record) when is_map(record) do
  record = record |> EctoStrategy.cast

  # ...build up and insert underlying component models based on attributes
  # in Location, including factory built or inserted associations.
end
```

Resolves #258


germsvel:
Hey @yordis, thanks for bringing this to my attention. @erikreedstrom thanks for opening this PR, and sorry it's taken so long to get a response. I've been meaning to look at this.  

It seems to me like `cast` is too much about the internals of ExMachina right now. And making this function public would be somewhat tangential to what  ExMachina is supposed to offer. 

I know #258 talks about a possible different approach, and it may be that we could go with something like that, though I'm still unsure about that being part of public api. I'm going to close this for now, and if you are interested we can continue the conversation in #258.

PR #248, opened 2017-09-08
Refactor Ecto.Strategy to use Ecto.Schema reflections

germsvel:
Summary
=======

`Ecto.Schema` has a `__schema__/1` function that allows us to
instrospect which fields are non-virtual, which are embeds, and which
are associations.

We already make use of this `__schema__` function for determining
associations and embeds, but the code seemed somewhat complicated in
that when we cast fields, we had to know to skip associations and
embeds by checking each fields type. By using `__schema__(:fields)` and `__schema__(:embeds)` we are able to cast fields first, then embeds, and finally associations.

This has the advantage of allowing us to untangle associations and
embeds code which, although is somewhat similar, does not overlap
completely (there are more conditions to check when casting
associations).

### DIFF
--- a/lib/ex_machina/ecto_strategy.ex
+++ b/lib/ex_machina/ecto_strategy.ex
@@ -29,36 +29,25 @@ defmodule ExMachina.EctoStrategy do
   defp cast(record) do
     record
     |> cast_all_fields
+    |> cast_all_embeds
     |> cast_all_assocs
   end
 
-  defp cast_all_fields(struct) do
-    struct
-    |> ExMachina.Ecto.drop_ecto_fields
-    |> Map.keys
-    |> cast_all_fields(struct)
-  end
+  defp cast_all_fields(%{__struct__: schema} = struct) do
+    field_keys = schema_fields(schema)
 
-  defp cast_all_fields(fields, struct) do
-    Enum.reduce(fields, struct, fn(field, struct) ->
-      casted_value = cast_field(field, struct)
-      Map.put(struct, field, casted_value)
+    Enum.reduce(field_keys, struct, fn(field_key, struct) ->
+      casted_value = cast_field(field_key, struct)
+
+      Map.put(struct, field_key, casted_value)
     end)
   end
 
-  defp cast_field(field, %{__struct__: schema} = struct) do
-    field_type = schema.__schema__(:type, field)
-    virtual_field? = !field_type
-    embed_type = schema.__schema__(:embed, field)
-    embedded_field? = !!embed_type
+  defp cast_field(field_key, %{__struct__: schema} = struct) do
+    field_type = schema.__schema__(:type, field_key)
+    value = Map.get(struct, field_key)
 
-    value = Map.get(struct, field)
-
-    if virtual_field? || embedded_field? do
-      value
-    else
-      cast_value(field_type, value, struct)
-    end
+    cast_value(field_type, value, struct)
   end
 
   defp cast_value(field_type, value, struct) do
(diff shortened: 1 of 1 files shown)


lib/ex_machina/ecto_strategy.ex
```diff
@@ -70,40 +59,76 @@ defmodule ExMachina.EctoStrategy do
...
-  defp cast_assoc(original_assoc, assoc_key, %{__struct__: schema} = struct) do
-    case original_assoc do
-      has_or_embeds_many when is_list(has_or_embeds_many) ->
-        Enum.map(has_or_embeds_many, &(cast_assoc(&1, assoc_key, struct)))
+  defp cast_embed(embeds_many, embed_key, struct) when is_list(embeds_many) do
+    Enum.map(embeds_many, &(cast_embed(&1, embed_key, struct)))
+  end
+  defp cast_embed(embed, embed_key, %{__struct__: schema}) do
+    case embed do
+      %{} ->
```
paulcsmith: Maybe this could be simplified using an `if`

```elixir
if embed
  # do stuff
end
```

lib/ex_machina/ecto_strategy.ex
```diff
@@ -70,40 +59,76 @@ defmodule ExMachina.EctoStrategy do
...
+        struct(assoc_type) |> Map.merge(assoc) |> cast()
 
       nil -> nil
     end
   end
 
-  defp get_schema_assocs(schema) do
-    schema.__schema__(:associations) ++ schema.__schema__(:embeds)
+  defp schema_fields(schema) do
+    schema_non_virtual_fields(schema)  -- schema_embeds(schema)
```
dbernheisel: extra space here before `--`

lib/ex_machina/ecto_strategy.ex
```diff
@@ -70,40 +59,76 @@ defmodule ExMachina.EctoStrategy do
...
+
+      nil -> nil
+    end
+  end
+
+  defp cast_all_assocs(%{__struct__: schema} = struct) do
+    assoc_keys = schema_associations(schema)
+
+    Enum.reduce(assoc_keys, struct, fn(assoc_key, struct) ->
+      casted_value = Map.get(struct, assoc_key) |> cast_assoc(assoc_key, struct)
```
dbernheisel: I see this in a couple places, but generally I think the accepted style is to use pipe chains when you can, 
like `struct |> Map.get(assoc_key) |> cast_assoc(assoc_key, struct)`

lib/ex_machina/ecto_strategy.ex
```diff
@@ -29,36 +29,25 @@ defmodule ExMachina.EctoStrategy do
...
-  defp cast_all_fields(struct) do
-    struct
-    |> ExMachina.Ecto.drop_ecto_fields
-    |> Map.keys
-    |> cast_all_fields(struct)
-  end
+  defp cast_all_fields(%{__struct__: schema} = struct) do
+    field_keys = schema_fields(schema)
+
+    Enum.reduce(field_keys, struct, fn(field_key, struct) ->
```
dbernheisel: it doesn't look like you use `field_keys` anywhere else, so stylistically you could do this:

```
schema
|> schema_fields
|> Enum.reduce(struct, fn(field_key, struct) -> 
...
```

lib/ex_machina/ecto_strategy.ex
```diff
@@ -70,40 +59,76 @@ defmodule ExMachina.EctoStrategy do
...
   end
 
-  defp cast_all_assocs(%{__struct__: schema} = struct) do
-    assocs = get_schema_assocs(schema)
+  defp cast_all_embeds(%{__struct__: schema} = struct) do
+    embed_keys = schema_embeds(schema)
 
-    Enum.reduce(assocs, struct, fn(assoc, struct) ->
-      casted_value = struct |> Map.get(assoc) |> cast_assoc(assoc, struct)
+    Enum.reduce(embed_keys, struct, fn(embed_key, struct) ->
```
dbernheisel: Same here about using a pipe chain 
```
schema
|> schema_embeds
|> Enum.reduce(struct, fn(embed_key, struct) ->
...
```

dbernheisel:
LGTM :+1: just have style comments.

lib/ex_machina/ecto_strategy.ex
germsvel: 👍 updated in b2ec4ca002be6d5a91609b45a4582e4652573e8d

lib/ex_machina/ecto_strategy.ex
germsvel: Nice catch. Updated in b2ec4ca002be6d5a91609b45a4582e4652573e8d

lib/ex_machina/ecto_strategy.ex
germsvel: This is a good catch. I tend to do this, but I think the preferred style is to have the first argument and then `|>`. I've updated this and the other mentions in b2ec4ca002be6d5a91609b45a4582e4652573e8d

germsvel:
Thanks @paulcsmith and @dbernheisel!

PR #253, opened 2017-10-22
add nil_params_for/2 to aid in writing @invalid_attrs

digitalcake:
If you follow the way the phoenix generators make tests you will see the use of @invalid_attrs where a map lists all of the keys but where all the values are nil. This helper allows for quickly replacing the the current use of @invalid_attrs.

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -156,7 +156,7 @@ defmodule ExMachina.Ecto do
   def nil_params_for(module, factory_name) do
     module
     |> params_for(factory_name, %{})
-    |> Enum.map( fn{k, _v} -> {k, nil} end )
+    |> Enum.map(fn{k, _v} -> {k, nil} end)
     |> Map.new
   end
 
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -27,6 +27,10 @@ defmodule ExMachina.Ecto do
           ExMachina.Ecto.string_params_for(__MODULE__, factory_name, attrs)
         end
 
+        def nil_params_for(factory_name) do
+          ExMachina.Ecto.nil_params_for(__MODULE__, factory_name)
+        end
+
         def params_with_assocs(factory_name, attrs \\ %{}) do
           ExMachina.Ecto.params_with_assocs(__MODULE__, factory_name, attrs)
         end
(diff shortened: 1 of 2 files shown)


lib/ex_machina/ecto.ex
```diff
@@ -135,6 +139,28 @@ defmodule ExMachina.Ecto do
...
+      # Returns %{name: => nil, admin: => nil}
+      nil_params_for(:user)
+  """
+  @callback nil_params_for(factory_name :: atom) :: %{optional(atom) => any}
+
+  @doc false
+  def nil_params_for(module, factory_name) do
+    module
+    |> params_for(factory_name, %{})
+    |> Enum.map( fn{k, _v} -> {k, nil} end )
```
pdawczak: I think the wrapping whitespace is redundant here

digitalcake:
I hope thats what you meant @pdawczak

pdawczak:
Thanks for addressing the comment.

I'm wondering if this specific case (building a struct with empty values) is actually worth extending the API? I'm even tempted to think this might even be _against_ the purpose of factory idea.

The main purpose of ex_machina (or factory girl) is to define *valid* entities, so in your specific testing cases, when using factory, you can override only some attributes (that you're focussing on in particular test), but you rely on the fact the rest of the attributes (not overridden ones) will still be valid.

Would you mind elaborate a bit more on why this might be useful?

digitalcake:
No prob, 

So I like to follow the style that phoenix generates for my test. 

IE: 
```
  describe "clients" do
    alias Beffect.Intake.Client

    @valid_attrs %{foo: "bar"}
    @update_attrs %{foo: "baz"}
    @invalid_attrs %{foo: nil}

```

I like to turn into 

```
  describe "clients" do
    alias Beffect.Intake.Client

    @valid_attrs params_for(:client)
    @update_attrs params_for(:client)
    @invalid_attrs params_for(:client) |> Enum.map(fn {k, _v} -> {k, nil} end) |> Map.new
```

Now with my helper I can simply 

```
  describe "clients" do
    alias Beffect.Intake.Client

    @valid_attrs params_for(:client)
    @update_attrs params_for(:client)
    @invalid_attrs nil_params_for(:client) 
```

digitalcake:
I will admit I just ran into a problem with module attributes like this as seen here 
https://github.com/thoughtbot/ex_machina/issues/254

So I no longer now if I like this style.

paulcsmith:
I think using private methods is more reliable and still keeps tests nice and clean. Maybe that would solve the problem just as well. What do you think?

As far as making invalid_params, I would do something like this

```elixir
params_for(:client, name: nil) # or some other thing that will make it invalid
```

Alternatively I believe you can add a function to do what this PR proposes, so we wouldn't need to extend ExMachina. 

```elixir
params_for(:client) |> nilify_values # Define `nilify_values` on your factory and import it
```

I think this is what I would prefer since it would not require new stuff in ExMachina

digitalcake:
thanks for the input @paulcsmith , I think that is a fair point.  Yeah I'm not so in love with this pr now that I find myself dealing with other issues related to assigning to the module attributes in the first place. I would be ok with no pulling this. 

Thanks again for you time :)

paulcsmith:
No problem. I'm happy to help. Good luck!

pdawczak:
As a sidenote @digitalcake, you could achieve similar effect with:

```elixir
Map.from_struct(%Beffect.Intake.Client{})
```

Obviously, this won't traverse associations, but for simple cases would do.

Hope that helps!

PR #257, opened 2017-10-28
Format code with Elixir 1.6 formatter

Arkham:
As title says 🎉


germsvel:
Hi @Arkham thanks so much for opening this PR. I'm going to hold off on doing this until I can set up a check in CI that ensures we're always doing this in new PRs.

nathany:
@germsvel You can use the `--check-formatted` option to `mix format` (if you have a `.formatter.exs` file).

Do note that new Elixir versions may change the formatting, which can cause test failures in CI.

Given the merge conflicts, I'd probably close this branch and format the current code on master instead. Thanks @Arkham.

germsvel:
Thanks @nathany! That is the plan. I'll hopefully get to that tomorrow. In the meantime, I agree it makes sense to close this. Thanks @Arkham for the PR! Sorry we didn't get to this

PR #266, opened 2017-12-11
Handle nil values for has_one associations

alco:
The has_one association allows for the referenced value to be nil.

ExMachina by default sets such fields to `%Ecto.Association.NotLoaded{}` when the field Is not mentioned in the factory definition. However, setting the field to nil explicitly in the factory definition caused a CaseClauseError in ExMachina.Ecto.handle_assoc.

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -216,6 +216,9 @@ defmodule ExMachina.Ecto do
         assoc = recursively_strip(original_assoc)
         Map.put(record, association_name, assoc)
 
+      nil ->
+        Map.put(record, association_name, nil)
+
       list when is_list(list) ->
         has_many_assoc = Enum.map(original_assoc, &recursively_strip/1)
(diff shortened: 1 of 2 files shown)


gmile:
Hi @germsvel, is there anything that blocks this PR? It would be great to have ability to explicitly set `nil` on association

germsvel:
@alco after giving this a second look, I don't think this will do what you want. Setting the value explicitly to `nil` will only allow the code execution to go further. But it will eventually `drop_fields_with_nil_values/1` in the `ExMachina.Ecto` module, and the association will be removed altogether. 

So if you wrote a test for this, 

```elixir
# this would fail
assert TestFactory.params_for(:user, best_article: nil) == %{name: "John Doe", articles: [], admin: false, best_article: nil}

# this would pass
assert TestFactory.params_for(:user, best_article: nil) == %{name: "John Doe", articles: [], admin: false}
```

PR #268, opened 2017-12-28
Clarify skipping application step of setup

azranel:
Just small clarification to adding ex_machina as application setup step


germsvel:
Thanks for helping us clarify this @azranel!

PR #275, opened 2018-03-06
Update underscore/1 to support Elixir 1.6's regex

unnawut:
`Regex.split/3`'s behavior was [changed in Elixir 1.6]( https://github.com/elixir-lang/elixir/issues/7385) and caused `ExMachina.Strategy.underscore/1` to return an extra underscore and broke all factory names that are generated by `ExMachina.Strategy.name_from_struct/1`.

This PR does ~2 things~ 1 thing:
- ~Fixes the regex inside `ExMachina.Strategy.underscore/1` to do negative lookbehind for a start of line.~
- ~Use a single `Regex.replace/4` instead of `Regex.split/3` + `Enum.join/2` which, [according to Jose Valim, is equivalent](https://github.com/elixir-lang/elixir/issues/7385#issuecomment-367969050).~
- Use [`Macro.underscore/1`](https://hexdocs.pm/elixir/Macro.html#underscore/1) instead of `Regex.split/3` + `Enum.join/2` to convert module name to snake case. (Thanks @vernomcrp for the suggestion!)

# Before

## Elixir 1.5.3 (success)

```elixir
$ elixir --version
Erlang/OTP 20 [erts-9.2.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

Elixir 1.5.3

$ mix test
The database for ExMachina.TestRepo has been dropped
The database for ExMachina.TestRepo has been created
................................................................

Finished in 0.3 seconds
64 tests, 0 failures

Randomized with seed 941966
```

## Elixir 1.6.2 (failed)

```elixir
$ elixir --version
Erlang/OTP 20 [erts-9.2.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

Elixir 1.6.2 (compiled with OTP 20)

$ mix test
The database for ExMachina.TestRepo has been dropped
The database for ExMachina.TestRepo has been created


  1) test name_from_struct returns the factory name based on passed in struct (ExMachina.StrategyTest)
     test/ex_machina/strategy_test.exs:23
     Assertion with == failed
     code:  assert ExMachina.Strategy.name_from_struct(%{__struct__: User}) == :user
     left:  :_user
     right: :user
     stacktrace:
       test/ex_machina/strategy_test.exs:24: (test)

...............................................................

Finished in 0.3 seconds
64 tests, 1 failure

Randomized with seed 568658
```

# After 

[...]

PR #276, opened 2018-03-13
Bump version to 2.2.0

germsvel:
Bump version to 2.2.0. Notable changes include,

- Support for lists in sequences (PR #227)
- Using `Macro.underscore/1` so that factory names don't break in elixir
1.6 (PR #275)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ complete changelog, see the git history for each version via the version links.
 
 [hex package page]: https://hex.pm/packages/ex_machina
 
+## [2.2.0]
+
+### Added
+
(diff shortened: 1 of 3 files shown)


dbernheisel:
LGTM. Verified that `mix.exs` already requires Elixir 1.2 which is required for `Macro.underscore`.

dbernheisel:
On second-thought, can we bump it to Elixir 1.4 (from Elixir 1.2). Looks like we use `List.pop_at` which was introduced to Elixir 1.4. I'll open a PR.

germsvel:
@dbernheisel that sounds good. I'll merge this in, and we can bump the elixir version in a separate PR.

PR #287, opened 2018-06-20
Remove obsolete `fields_for/2` function

### LINKED ISSUE #157: use ExMachina.Ecto defines functions with no local return
In `test/support/factory.ex`, I have

```
  use ExMachina.Ecto., repo: MyApp.Repo
```

This works for the factories just fine, but `mix dialyze` thinks all function defined by the `use` have no local return:

```
test/support/factory.ex:22: Function create/1 has no local return

test/support/factory.ex:22: Function create/2 has no local return

test/support/factory.ex:22: Function create_list/3 has no local return

test/support/factory.ex:22: Function create_pair/2 has no local return

test/support/factory.ex:22: Function fields_for/1 has no local return
```

My work-around is 

```
 @dialyzer {:no_return, create: 1, create: 2, create_list: 3, create_pair: 2, fields_for: 1}
```

Since the functions are only there to raise errors, their specs can say as much as the work-around would not be necessary.  I'll open a PR to do that.

moxley:
Fixes #157

Since version 1.0.0, when users of ex_machina try to call the deprecated `fields_for/2` function, it raises an exception. This doesn't seem to provide any benefit. If instead it's removed, calls to it will fail at compile time. But if it stays in, calls will fail at runtime, possibly in production.

This PR solves the Dialyzer issue and it provides a better solution for dealing with the obsolete function.

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -34,10 +34,6 @@ defmodule ExMachina.Ecto do
         def string_params_with_assocs(factory_name, attrs \\ %{}) do
           ExMachina.Ecto.string_params_with_assocs(__MODULE__, factory_name, attrs)
         end
-
-        def fields_for(factory_name, attrs \\ %{}) do
-          raise "fields_for/2 has been renamed to params_for/2."
-        end
       end
     else
       raise ArgumentError,


sbrink:
+1

tielur:
Would this fix these types of dialyzer issues?

```elixir
________________________________________________________________________________
test/support/factory.ex:11:no_return
Function subject_factory/0 has no local return.
________________________________________________________________________________
test/support/factory.ex:3:no_return
Function fields_for/1 has no local return.
________________________________________________________________________________
test/support/factory.ex:8:no_return
Function patient_factory/0 has no local return.
________________________________________________________________________________
```

moxley:
@tielur: Possibly. Try pulling in this branch into your mix.exs `deps()` to find out.

germsvel:
I think this makes sense to remove, especially since the deprecation was introduced in 1.0.0 and we're at 2+. Thanks @moxley for providing some info in the PR description!

moxley:
❤️

PR #288, opened 2018-07-13
Upgrade to CircleCI 2

germsvel:
From CircleCi,

screen shot 2018-07-13 at 4 22 31 pm

So we add CircleCI 2 config file that is as close as possible to our current setup.


.circleci/config.yml
```diff
@@ -0,0 +1,53 @@
...
+      - save_cache:
+          key: v1-build-cache-{{ .Branch }}
+          paths: "_build"
+      - save_cache:
+          key: v1-build-cache
+          paths: "_build"
+
+      - run:
+          name: Wait for DB
+          command: dockerize -wait tcp://localhost:5432 -timeout 1m
```
SViccari: @germsvel I realize I'm late to the review party but I find this line interesting. Were you running into an issue where the tests were running before the docker services were ready?

germsvel: @SViccari I wasn't running into that issue. I've putting this in circle configs because I got it from https://circleci.com/docs/2.0/language-elixir/. The example there says the following, 

```yml
      - run:  # special utility that stalls main process until DB is ready
          name: Wait for DB
          command: dockerize -wait tcp://localhost:5432 -timeout 1m
```

PR #308, opened 2018-11-05
Format code with mix format

germsvel:
Format all code with `mix format`, and set `--check-formatted` in CI to ensure we keep the code formatted from now on.


germsvel:
Seems our version of elixir is a bit old. I'll close this until we update that.

PR #327, opened 2019-02-01
Add support Ecto struct for params_for

Nitrino:
Added support Ecto structs for the `params_for` and `string_params_for` functions.
It allows use build chains with already built factories and reuse modification functions.
e.g.:
```elixir
build(:user) |> with_password |> params_for
```


Nitrino:
Hi guys. What about this feature?

germsvel:
Hi @Nitrino , sorry for not responding sooner. Somehow I missed this PR being open. I'm trying to understand the motivation behind this. 

Instead of doing this: 

```elixir
build(:user) |> with_password() |> params_for()
```

Couldn't you just use this? 

```elixir
params_for(:user) |> with_password()
```

Nitrino:
Hi @germsvel, thanks for the answer. 
I can hardly remember the reason, but I think the idea was to build the structure once and then use it both as an instance and for parameters.

If we use this way:
```elixir
user = params_for(:user) |> with_password()
```
We will not be able to use the `user` variable for insert
```elixir
params_for(:user) |> with_password() |> insert()
** (ArgumentError) %{admin: false, password: "", name: "John Doe"} is not an Ecto model. Use `build` instead
```
Thus, in some cases, we will have to create a `build_user` for insert and `user_params` for pass params, instead to create one `build_user` and call ` params_for`

```elixir
build_user = build(:user) |> with_password |> ... |> ...
user_params = params_for(:user) |> with_password  |> ... |> ...

# This can be simply:
build_user = build(:user) |> with_password |> ... |> ...
user_params = build_user |> params_for()
```

We will also have to to get away from type matching in `with_password`  and other function modifiers or write second same function with different headers.
```elixir
def with_password(%MyApp.User{} = post)
def with_password(%{} = post)
```

germsvel:
Thanks for responding! I can't blame you for not remembering the motivation since this has been open for a little while.

I see the examples you give, but I'm still not convinced this is functionality that makes sense to add.  But that may be because I don't understand when we would use it. 

---

You mention:

> Thus, in some cases, we will have to create a build_user for insert and user_params for pass params, instead to create one build_user and call params_for
```elixir
build_user = build(:user) |> with_password |> ... |> ...
user_params = params_for(:user) |> with_password  |> ... |> ...

# This can be simply:
build_user = build(:user) |> with_password |> ... |> ...
user_params = build_user |> params_for()
```


I don't see what scenario we'd be trying to reuse the built user, but I may just not be thinking of the right use case. To give you an example, I usually use `params_for` when I'm trying to test updating a record. So I might insert a user, and then generate some user params that are different from the existing user to make sure the update works. For example,

```elixir
user = insert(:user)
params = params_for(:user, email: "randomEmail@example.com")

# update user here with params

assert updated_user.email == params["email"]
```

But in those scenarios, I don't really want to reuse the `user` struct. If we really want to reuse it, could we use [`Map.from_struct`](https://hexdocs.pm/elixir/Map.html#from_struct/1) to get the values: 

```elixir
user = insert(:user)
params = user |> Map.from_struct() |> Map.put(:email, "hello@example.com")

# do the rest
````

---

You also mention: 

> We will also have to to get away from type matching in with_password and other function modifiers or write second same function with different headers.

```elixir
def with_password(%MyApp.User{} = post)
def with_password(%{} = post)
```

I don't think it's necessarily bad that you can pattern match on each of those. Perhaps you want them to behave slightly different (e.g. when just a map you don't care about hashing the password).

But I think we could also remove the restriction and do something like this, right? 

```elixir
def with_password(post) do
  %{post | password: "random password"}
end

# or

def with_password(post) do
  Map.put(post, :password, "password")
end
```

I know it's been a while, so the exact use case is hard to remember. But if you think of a scenario where you would use the `build(:user) |> params_for()`, let me know! I'm happy to keep chatting. I just don't want to introduce code that doesn't have a clear use case.

Nitrino:
@germsvel Thanks for the detailed answer.
I agree with your arguments and now I can not come up with cases of using this.
Let's close this pull request and if the need arises, I will return with real use cases for discussion

PR #331, opened 2019-02-07
Add reset/1 with the option to reset specific sequences at once

ArielFarias:
Fixed docs and create `reset/1` which can receive a list with sequence names or a single sequence name. This will reset to 0 the sequence names in the list and leave any other sequence with the current index.


germsvel:
Thanks for these changes @ArielFarias. They look good!

PR #342, opened 2019-04-06
Fix indentation for doc example

lccezinha:
Hey guys, just a simple indentation adjust in the documentation for `merge_attributes/2` function.

Before:

before

After:

after


germsvel:
@lccezinha thank you!

Issue #369, opened 2019-11-15
`string_params_for` only available when using ExMachina.Ecto


col:
I often use ExMachina in projects that don't use Ecto but I'd still like to use the `string_params_for` helper. Currently it's only available if I use Exmachina.Ecto.

Why is `string_params_for` dependant on Ecto? 

Also, the docs (README) are currently wrong as they specify that `params_for` and `params_with_assocs` require ExMachina.Ecto but don't mention it for `string_params_for`.

col:
Closing this as it's essentially a duplicate of https://github.com/thoughtbot/ex_machina/issues/355

Issue #355, opened 2019-06-27
Proposal: Allow using ExMachina.Ecto without a repo.


bamorim:
Ecto has been splitting the persistence layer a lot lately. As of Ecto 3, `:ecto_sql` is even a different package.

One of the usages of Ecto is just using plain embedded schemas to define some fields. I personally use them a lot in our web layer.
In our most recent project, every mutation method on our controllers are accompanied by an Ecto embedded schema. Our web layer doesn't have any Repository, since it doesn't have any normal schemas, just embedded schemas.

That being said, I don't think `ExMachina.Ecto` should enforce having a `:repo`.

My sugesstion is that if no repo is passed, it defines all methods except the insert ones.
The `params_for` and `string_params_for` are useful even without the Repo (since it strips Ecto related stuff).

So my suggestion is simple (I can open a PR for that easily): just allow repo to not be passed.

It would look something like

```elixir
defmacro __using__(opts) do
  verify_ecto_dep()

  repo = Keyword.get(opts, :repo)

  for_ecto_with_repo = quote do
    use ExMachina.EctoStrategy, repo: unquote(repo)

    def params_with_assocs(factory_name, attrs \\ %{}) do
      ExMachina.Ecto.params_with_assocs(__MODULE__, factory_name, attrs)
    end

    def string_params_with_assocs(factory_name, attrs \\ %{}) do
      ExMachina.Ecto.string_params_with_assocs(__MODULE__, factory_name, attrs)
    end
  end

  for_all = quote do
    use ExMachina

    def params_for(factory_name, attrs \\ %{}) do
      ExMachina.Ecto.params_for(__MODULE__, factory_name, attrs)
    end

    def string_params_for(factory_name, attrs \\ %{}) do
      ExMachina.Ecto.string_params_for(__MODULE__, factory_name, attrs)
    end
  end

  if not is_nil(repo) do
    quote do
      unquote(for_all)
      unquote(for_ecto_with_repo)
    end
  else
    for_all
  end
end
```

What do you guys think?

johantell:
I think using `false` over `nil` could make it more explicit (otherwise I can see cases where one could forgot to add/misspell the `repo` option and get a lot of headaches)

```elixir
use ExMachina.Ecto, repo: false
```

germsvel:
[...]

PR #370, opened 2019-11-17
Allow ExMachina.Ecto to be used without the :repo option

col:
We should be able to use Ecto without a repo. This is useful when using `ecto` without `ecto_sql`.

ExMachina currently raising an ArgumentError when using ExMachina.Ecto without a repo. This change now allows the repo to be nil and raises a RuntimeError when the insert/1 function is called instead.

New error message should still be helpful enough for people that simply forgot to add the :repo option.

```
insert/1 is not available unless you provide the :repo option. Example:

```

I believe this addresses the following issues:
https://github.com/thoughtbot/ex_machina/issues/369
https://github.com/thoughtbot/ex_machina/issues/355

### DIFF
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -15,34 +15,25 @@ defmodule ExMachina.Ecto do
   defmacro __using__(opts) do
     verify_ecto_dep()
 
-    if repo = Keyword.get(opts, :repo) do
-      quote do
-        use ExMachina
-        use ExMachina.EctoStrategy, repo: unquote(repo)
+    quote do
+      use ExMachina
+      use ExMachina.EctoStrategy, repo: unquote(Keyword.get(opts, :repo))
 
-        def params_for(factory_name, attrs \\ %{}) do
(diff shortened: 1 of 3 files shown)


germsvel:
@col thanks so much for opening up this PR! It looks really good. 

I saw you added some tests for when we raise the error. Would you also be willing to add a couple of tests that ensure `params_for` and `string_params_for` work without the `repo` being passed in?

col:
Sure will do.

col:
@germsvel - Done. I kept the tests pretty basic. Let me know if you'd like the tests to cover anything else more specifically. Looks like the repo is only used by the `ExMachina.EctoStrategy` which only affects the `insert` functions.

germsvel:
Thanks @col! I gave it a quick glance, and things look good. I'll give it a better look when I have a few more minutes, but I think this is probably good to merge.

germsvel:
This looks good! Thanks for the PR @col! 🎉

col:
> This looks good! Thanks for the PR @col! 🎉

You're welcome. Thanks for merging. :+1:

Issue #280, opened 2018-04-10
Missing multi-tenancy support


joeppeeters:
Hi, 

In our project we implement multi-tenancy by setting prefixes on the `Repo` (eg. `MyApp.Repo.insert(%User{name: 'John'}, prefix: tenant_id)`

We'd like to reuse this behaviour in our tests, but ex_machina doesn't seem to support it currently. Would that be something you would consider to add?

An easy way would be to create a factory for a test tenant by configure it in the `use` statement:
```
defmodule MyFactory do
  use ExMachina.Ecto, repo: MyRepo, repo_options: [prefix: "test_tenant"]

  ...
end
```
and then rewriting the `ExMachina.EctoStrategy` to:
```
...
def handle_insert(%{__meta__: %{__struct__: Ecto.Schema.Metadata}} = record, %{repo: repo, repo_options}) do
    record
    |> cast
    |> repo.insert!(repo_options)
  end
...
```
or more flexible at runtime by introducing an `insert/3` method.

Is this something worth to consider? Or do you see any other ways to achieve the same behaviour?

churcho:
My team just hit this snag. 
What options are there with regards to this?

germsvel:
Thanks for posting here @churcho, and sorry @joeppeeters for taking so long to respond here! We currently don't have anything that supports multitenancy. 

There is also a conversation on https://github.com/thoughtbot/ex_machina/issues/273#issuecomment-367823475 that might be related, since they're also trying to provide extra arguments to `Repo.insert!`. I think the only option for now is to create a custom strategy that has those options defined in `handle_insert`. Naturally, that's not ideal because you _can't_ really defined them at runtime like this, 

```elixir
build(:user) |> insert(prefix: "test_tenant")
```

You would have to define them in each strategy you create, 

```elixir
# in TestTenantEctoStrategy or something like that
def handle_insert(%{__meta__: %{__struct__: Ecto.Schema.Metadata}} = record, %{repo: repo}) do
    record
    |> cast
    |> repo.insert!(prefix: "test_tenant")
  end
```

Is that something that might work?

bokay:
Hi @germsvel 
I have tried what you suggest with a custom strategy:

```
defmodule Datastore.TenantEctoStrategy do
  use ExMachina.Strategy, function_name: :insert
...
end
```

However, this leads to a conflict with the `insert` function defined within the library Ecto strategy:

```
== Compilation error in file test/support/factory.ex ==
[...]

PR #377, opened 2020-04-25
Pass repo_options to repo.insert! to support TenantFactories

joeppeeters:
Closing #280

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,6 +370,24 @@ change this line in `mix.exs`:
 defp elixirc_paths(:test), do: ["lib", "web", "test/support", "test/factories"]
 ```
 
+## Multi-tenant support
+
+If you are depending on query prefixes to switch between tenant schemas you can 
+pass `repo_options` to your factory which can be used to set the `:prefix` on your insert queries.
+This allows you to create a factory per tenant.
+
+```elixir
+defmodule MyApp.TenantFactory
+  use ExMachina.Ecto, repo: MyApp.Repo, repo_options: [prefix: "test_tenant"]
+
+  def user_factory do
+    ...
+  end
+
+  ...
+end
+```
+
 ## Custom Strategies
 
 You can use ExMachina without Ecto, by using just the `build` functions, or you
--- a/test/ex_machina/ecto_test.exs
+++ b/test/ex_machina/ecto_test.exs
@@ -36,6 +36,22 @@ defmodule ExMachina.EctoTest do
       end
     end
 
+    defmodule MockRepo do
+      def insert!(_model, opts), do: opts
+    end
+
+    defmodule MockedTenantFactory do
+      use ExMachina.Ecto, repo: MockRepo, repo_options: [prefix: "test_tenant"]
+
+      def user_factory do
+        %ExMachina.User{}
+      end
+    end
+
+    test "it passes repo_options to Repo.insert!" do
+      assert [prefix: "test_tenant"] == MockedTenantFactory.insert(:user)
+    end
+
     test "params_for/1 still works as expected" do
       user_params = NoRepoTestFactory.params_for(:user)
 
--- a/lib/ex_machina/ecto.ex
+++ b/lib/ex_machina/ecto.ex
@@ -17,7 +17,9 @@ defmodule ExMachina.Ecto do
 
     quote do
       use ExMachina
-      use ExMachina.EctoStrategy, repo: unquote(Keyword.get(opts, :repo))
+      use ExMachina.EctoStrategy,
+        repo: unquote(Keyword.get(opts, :repo)),
+        repo_options: unquote(Keyword.get(opts, :repo_options, []))
 
       def params_for(factory_name, attrs \\ %{}) do
         ExMachina.Ecto.params_for(__MODULE__, factory_name, attrs)
--- a/lib/ex_machina/ecto_strategy.ex
+++ b/lib/ex_machina/ecto_strategy.ex
@@ -20,10 +20,10 @@ defmodule ExMachina.EctoStrategy do
     """
   end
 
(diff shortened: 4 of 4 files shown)


test/ex_machina/ecto_test.exs
```diff
@@ -36,6 +36,22 @@ defmodule ExMachina.EctoTest do
...
+    defmodule MockedTenantFactory do
+      use ExMachina.Ecto, repo: MockRepo, repo_options: [prefix: "test_tenant"]
+
+      def user_factory do
+        %ExMachina.User{}
+      end
+    end
+
+    test "it passes repo_options to Repo.insert!" do
+      assert [prefix: "test_tenant"] == MockedTenantFactory.insert(:user)
```
germsvel, line 52: Thanks for adding a test with this 🎉 

I see that we're using the `MockRepo` defined above which only returns the `opts` back (which makes for really easy testing). But I'm wondering, how do you feel about adding a test that exercises the full `insert` into a separate database? Since it's a new feature, it would be nice to have more certainty that everything works as expected. What do you think?

README.md
```diff
@@ -370,6 +370,24 @@ change this line in `mix.exs`:
...
+defmodule MyApp.TenantFactory
+  use ExMachina.Ecto, repo: MyApp.Repo, repo_options: [prefix: "test_tenant"]
+
+  def user_factory do
+    ...
+  end
+
+  ...
+end
+```
```
germsvel, line 389: 🎉

test/ex_machina/ecto_test.exs
joeppeeters: Mocking the repo was done in an attempt to prevent that, but I agree this changing the return spec of `insert/1` makes it a bit weird. 

I'm fine with creating an integration test to the database. Any advise on how to set that up? I'll try to look into it later this week.

germsvel: I haven't looked deeply into how to set the multi-tenancy with Ecto either. I think all the setup for Ecto in ExMachina's tests comes from `test/support/test_repo.ex` and `config/test.exs`. 

If I have time, I will also try to give this a shot, but I'm a bit swamped right now. Thank you for getting this PR this far 🎉. Let's see if someone else comes along and can finish it or if either of us frees up.

mojidabckuu: @germsvel So to merge this PR is all about to add a sample test with multi tenancy? The sample `User.ex` schema would be enough?

germsvel: > So to merge this PR is all about to add a sample test with multi tenancy?

Yes, I think if we could get a full integration test with multi-tenancy we'd be good. I just want to make sure this works correctly and that we have it tested. 

> The sample User.ex schema would be enough?

I don't know if you mean the `User` schema defined above this test or something like https://github.com/thoughtbot/ex_machina/blob/master/test/support/models/user.ex. I think ideally we have a schema that has a couple of fields instead of the one defined above

mojidabckuu: @germsvel ok, copy that! I propose to create another one `TenantUser` and put a `:nickname, :text`, then create same db with different prefixes and ensure that inserted tenant user exists after `insert(:tenant_user` call. Would it be enough to cover the feature?

suan: @germsvel Any thoughts on @mojidabckuu's last comment? We too are in need of this feature! 🙂

germsvel: Sorry, I missed that. Yes, that sounds like a good test to me!

README.md
zorn: Found this while looking for tenet-support fixtures for my own project. If this PR still needs more test, and the above described acceptance criteria are still valid, let me know and maybe I can take a shot. Love using this on my other projects.

test/ex_machina/ecto_test.exs
zorn: Found this while looking for tenet-support fixtures for my own project. If this PR still needs more test, and the above described acceptance criteria are still valid, let me know and maybe I can take a shot. Love using this on my other projects.

germsvel: Thanks for jumping in @zorn. Yeah, the above acceptance criteria is still what's needed. I just want to make sure we have a test testing multi-tenancy. I'd love to get this merged after that. Let me know if you need any help or if that's good enough to move forward.

zorn: I worked on this today but ran into the same issues I ran into a year ago when trying to use the Ecto sandbox env and dynamic tenants. 

Specifically in a traditional test, the database is dropped/created/migrated before the tests run inside a db transaction (to allow for fast async tests), but when doing a dynamic tenant it's not obvious to me how to run prefixed migrations inside a the transaction-based checkouts after manually making the tenant with something like `Repo.query("CREATE SCHEMA test_tenant")`. I saw `Ecto.Migrator.run(Repo, :up, options)` and `Ecto.Adapters.SQL.Sandbox.mode(Repo, :auto)` which are related but I keep running into pid ownership issues no matter what I try.

The other thing to keep in mind (for anyone who picks this up) is that Postgres and MySQL do this behavior differently. In one of my projects I've leaned on [Triplex](https://github.com/ateliware/triplex) for this abstraction. Not sure if this library currently or wants to support MySQL.

With what I know now, on an individual project, it wouldn't be too hard to code a project's `mix test` to do a prefix migration with `mix ecto.migrate --prefix "prefix_1"` ahead of running tests and making some assumptions about a tenant name. Sadly, dynamic tenants seem to be much harder for me to get working, with or without this library. :(

germsvel:
I just opened https://github.com/thoughtbot/ex_machina/pull/411, which might help with the multi-tenancy issue at the point of calling `insert/3`. I think it makes sense to also have a more global `repo_options` when we `use ExMachina.Ecto`, so I think this PR would still be desired.

PR #384, opened 2020-06-18
Update Reference to Prior Art

jherdman:
FactoryGirl hasn't been named as such for quite a while. I've preserved the name in the commit to ensure this library is still found when searching for something along the lines of "elixir factorygirl".


germsvel:
Thank you for making this change! Even learned a new word:

> née: originally or formerly called

PR #399, opened 2020-10-21
Fix Elixir 1.11 compilation warning

### LINKED ISSUE #398: Compiler warning when using an a project using Elixir 1.11
Running `mix compile` in a project using Elixir 1.11 causes a compiler error which originates in `ex_machina`. The `ExMachina.Strategy.__using__` macro has a function call without parentheses. The trouble is that even when running `mix compile --warnings-as-errors --no-deps-check`, the compilation still fails. I assume this is because the `use ExMachina.Strategy` call compiles that function into our own code, so the compiler thinks that error is coming from our code, rather than a library.

_Edit_ screenshot
Screen Shot 2020-10-19 at 2 18 54 PM

crbelaus:
This PR fixes the compilation warning produced by Ex Machina under Elixir 1.11.

Closes #398

### DIFF
--- a/lib/ex_machina/strategy.ex
+++ b/lib/ex_machina/strategy.ex
@@ -52,7 +52,7 @@ defmodule ExMachina.Strategy do
 
       defmacro __using__(opts) do
         custom_strategy_module = __MODULE__
-        function_name = custom_strategy_module.function_name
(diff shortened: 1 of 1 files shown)


germsvel (approved):
Thanks for opening this fix! 🎉

Issue #402, opened 2020-11-11
Proposal: Introduce build_lazy/2


germsvel:
The problem
------------

Many people build factories like this and run into uniqueness issues or foreign key issues:

```elixir
insert_pair(:account, email: build(:email))
```

They think that each account will have a different email (especially if they set a sequence). The problem is that this is just function calling. So the above is equivalent to this:

```elixir
email = build(:email)
insert_pair(:account, email: email)
```

In other words, people are trying to insert multiple accounts with the _same_ email. They'll get a validation error if they have uniqueness.

Proposed Solution
------------------

The solution is to somehow delay evaluation of the `email` factory. We can do this in one of two ways: 

### Anonymous function

The simplest way to do this is to capture it in an anonymous function, basically doing this:

```elixir
build_pair(:account, email: fn -> build(:email) end)
```

We could wrap that in a new function called `build_lazy/2` (or possibly a different name), so that `build_lazy/2` returns `fn -> build(:factory_name) end`

The trick then lies in the `build/2` function. It now checks all attributes passed to it. If it's a function, it evaluates it. That way, when we're building a list of things, we evaluate the `build(:email)` once per account we're creating.

I built a proof-of-concept of this approach in https://github.com/thoughtbot/ex_machina/commit/c9816d245ded9ed5a54c4977cb2b3f27ef7ee418. 

### Private data structure

A more advanced implementation might create a struct that we can pattern match against when executing later.

So `build_lazy/2` would return a data structure with all the information needed to execute the function later. The `build/2` function would then perform a similar role as it does in the anonymous function case: it would pattern match on the custom struct and evaluate the "lazy" factory at the time when we're iterating through the attributes. 

This is a proof-of-concept changing the anonymous function implementation to a struct-based implementation: https://github.com/thoughtbot/ex_machina/commit/afc0cb2d01040f6df56653c7e3717ccbf95bef75. 

### Trade-offs

The anonymous function seems simpler to me, and I like that people could even use anonymous functions to delay execution instead of using `build_lazy/2`. But I don't know if that would be a breaking change. If people have been using anonymous functions in their factories as an attribute like this: 

```elixir
def user_factory do
  %{
    name: "Some Name",
    creation_callback: fn -> User.user_created!() end}
  }
end
```

[...]

Issue #385, opened 2020-06-23
Uniq column error


Adzz:
I have this factory:

```elixir
def thing_factory() do
  %Thing{name: Ecto.UUID.generate()}
end
```
name is a column with a unique index.

When I do this:
```elixir
for _ <- 1..5 do
  thing = build(:thing)
  insert_list(4, :other_thing, thing: thing)
end
```

I get:
```sh
 ** (Ecto.ConstraintError) constraint error when attempting to insert struct:
 * thing_name_index (unique_constraint)
```

essentially saying the name is not uniq. If I change it to this it works:
```elixir
for _ <- 1..5 do
  thing = insert(:thing)
  insert_list(4, :other_thing, thing: thing)
end
```

Am I doing something wrong?

germsvel:
Hi @Adzz, I'm not a 100% sure, but I think this is what's happening. 

First, I think the  `for` comprehension is irrelevant to the problem. So think doing this will still trigger the error: 

```elixir
thing = build(:thing)
insert_list(4, :other_thing, thing: thing)
```

`build(:thing)` just creates a struct with the auto-generated `UUID`. You're then trying to insert 4 `:other_thing`, each of which has an associated (but not inserted) `thing` with the same name. 

```
# insert :other_thing_1 (and insert :thing with auto generated UUID) => succeeds
# insert :other_thing_2 (and attempt to insert :thing with same UUID) => fails b/c that UUID is already in the database. 
# :other_thing 3 and 4 aren't even attempted
```

If you `insert(:thing)` instead of just `build(:thing)` to begin with, `insert_list(:other_thing)` will not attempt to insert `thing` when it's recursively inserting those 4 records. 

Does that make sense? I hope this helps!

germsvel:
Opened up a PR that I think fixes this issue. Let me know what you think -> https://github.com/thoughtbot/ex_machina/pull/406

Issue #373, opened 2019-12-28
Constraint error when attempting to insert list with a build association (with sequence)


josephan:
Consider a chat application with the following schema (users table has a unique index on emails column):

![download](https://user-images.githubusercontent.com/8878596/71538661-b0205500-28fc-11ea-9bb1-3922a7352871.png)

When I want to insert a list of `chat_memberships` like so:
```elixir
    chat_room = insert(:chat_room)

    insert_list(
      20,
      :chat_membership,
      chat_room: chat_room,
      user: build(:user)
    )
```
I get the following error, even thought the `user` factory has a sequence for the `email` field:
```
     ** (Ecto.ConstraintError) constraint error when attempting to insert struct:

         * users_email_index (unique_constraint)

     If you would like to stop this constraint violation from raising an
     exception and instead add it as an error to your changeset, please
     call `unique_constraint/3` on your changeset with the constraint
     `:name` as an option.

     The changeset has not defined any constraint.
```

However if I setup my data like this instead, it works:
```elixir
    chat_room = insert(:chat_room)

    users = build_list(20, :user)

    Enum.each(users, fn user ->
      insert(:chat_membership, user: user, chat_room: chat_room)
    end)
```

Is this the intended behaviour?

I've created a bare minimum phoenix app to reproduce this issue: https://github.com/josephan/ex_machina_sample

Factory file:
https://github.com/josephan/ex_machina_sample/blob/master/test/support/factory.ex

The test file that demonstrates the issue: https://github.com/josephan/ex_machina_sample/blob/master/test/ex_machina_sample_web/sample_test.exs

In this app there is a `users` table and a `chat_rooms` table.
These tables have a many to many relationship with `chat_memberships` table.

germsvel:
Hi @josephan thanks so much for the detailed description along with an app to reproduce the issue! I can't begin to tell you how incredibly helpful that is. I haven't had the time to take a look at the app yet, but I think I know what might be going on. If it's not it, I'll take a look at if for sure when I have few more minutes. 

Looking at the description you posted, I think the issue comes from the fact that factories are built eagerly. So when you call `build(:user)` we're building the `%User{}` struct immediately. In other words, in the first scenario, this is what is happening:

```elixir
# this
    insert_list(
      20,
      :chat_membership,
      chat_room: chat_room,
      user: build(:user)
    )

# is equivalent to this
    insert_list(
      20,
      :chat_membership,
      chat_room: chat_room,
[...]

PR #406, opened 2020-12-08
Introduce `build_lazy/2`

germsvel:
Closes https://github.com/thoughtbot/ex_machina/issues/402, https://github.com/thoughtbot/ex_machina/issues/385, and https://github.com/thoughtbot/ex_machina/issues/373

The problem
------------

Consider this scenario. People want a different email per account (especially if the email factory has a sequence), but they do this:

```elixir
build_pair(:account, email: build(:email))
```

The problem is that `build/2` is just function calling. So the above is equivalent to this:

```elixir
email = build(:email)
build_pair(:account, email: email)
```

In other words, we get one email factory for all of the accounts.

The problem is worse when using it with Ecto. We can image the following scenario:

```elixir
insert_pair(:account, user: build(:user))
```

If the user factory has a uniqueness constraint, `insert_pair/2` will raise an error because we'll try to insert a user with the same value (even if using a sequence).

Solution
--------

The solution is to delay evaluation of the factory. We do this by introducing a private struct `%ExMachina.InstanceTemplate{}` to hold the data necessary to build the instance of that factory.

Note that this struct is private and liable to change, so users of the library shouldn't depend on it.

The trick then lies in the `build/2` function. We make it a terminal function in that it will evaluate any lazy factories recursively. To do that, we update the `build/2` function to evaluate `ExMachina.InstanceTemplate` structs after building the `build/2` factory.

In an early implementation of this feature, I evaluated the lazy factories when we first got the attributes in `build/2` (i.e. `attrs |> evaluate_lazy_factories() |> Enum.into(%{})`). But then I opted for evaluating the lazy factories after the factory is created because that allows us to use `build_lazy/2` from within factory definitions, and not just when combined with `build/2` or `build_*` functions.

### DIFF
--- a/lib/ex_machina.ex
+++ b/lib/ex_machina.ex
@@ -42,6 +42,10 @@ defmodule ExMachina do
         ExMachina.build(__MODULE__, factory_name, attrs)
       end
 
+      def build_lazy(factory_name, attrs \\ %{}) do
+        ExMachina.build_lazy(__MODULE__, factory_name, attrs)
+      end
+
       def build_pair(factory_name, attrs \\ %{}) do
         ExMachina.build_pair(__MODULE__, factory_name, attrs)
       end
@@ -165,8 +169,16 @@ defmodule ExMachina do
 
 
   If you want full control over the factory attributes, you can define the
-  factory with `[factory_name]_factory/1`. Note that you will need to merge the
-  attributes passed if you want to emulate ExMachina's default behavior.
+  factory with `[factory_name]_factory/1`.
+
+  Caveats:
+
+  - ExMachina will no longer merge the attributes for your factory. If you want
+  to do that, you can merge the attributes with the `merge_attributes/2` helper.
+
+  - You cannot use `c:build_lazy/2` within the factory definition, since ExMachina
+  will no longer perform it's usual evaluation of lazy factories after the
+  factory creation.
 
   ## Example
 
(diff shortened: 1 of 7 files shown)


lib/ex_machina.ex
```diff
@@ -221,6 +236,39 @@ defmodule ExMachina do
     Code.ensure_loaded?(module) && function_exported?(module, function_name, 0)
   end
 
+  @doc """
+  Builds a factory that will be evaluated when building another factory's
```
germsvel, line 240: Would love thoughts on the documentation

priv/test_repo/migrations/1_migrate_all.exs
```diff
@@ -3,23 +3,29 @@ defmodule ExMachina.TestRepo.Migrations.MigrateAll do
...
+      add(:visits, :decimal)
     end
 
     create table(:comments) do
-      add :article_id, :integer
-      add :author, :map
-      add :links, {:array, :map}, default: []
+      add(:article_id, :integer)
+      add(:author, :map)
+      add(:links, {:array, :map}, default: [])
```
germsvel, line 28: Autoformatter added all the parentheses.

priv/test_repo/migrations/1_migrate_all.exs
```diff
@@ -3,23 +3,29 @@ defmodule ExMachina.TestRepo.Migrations.MigrateAll do
...
+      add(:name, :string)
+      add(:admin, :boolean)
+      add(:net_worth, :decimal)
     end
 
+    create table(:publishers) do
+      add(:pub_number, :string)
+    end
+
+    create(unique_index(:publishers, [:pub_number]))
```
germsvel, line 15: Added this uniqueness constraint to properly test the scenario in ecto where `insert_pair(:account, publisher: build(:publisher))` would raise a constraint error.

WilHall (approved):
This implementation makes sense to me, though I am curious what the benefits and drawbacks of this approach are over allowing factory attributes to accept functions? This would remove the need for a separate `build_lazy` method which would then become:

```elixir
insert_pair(:account, user: fn -> build(:user) end)
```

This would be a bit more general purpose, allowing other types of delayed execution when resolving attributes, and would potentially allow for passing context into the anonymous function if we ever needed to do that.

germsvel:
@WilHall that's an excellent question. I wrote a little bit about it in #402, but I should've mentioned it here. 

The anonymous function seems simpler to me, and I like that people could even use anonymous functions to delay execution instead of using build_lazy/2. **But I don't know if that would be a breaking change.** If people have been using anonymous functions in their factories as an attribute like this:

```elixir
def user_factory do
  %{
    name: "Some Name",
    creation_callback: fn -> User.user_created!() end}
  }
end
```

We would all of the sudden evaluate those anonymous functions when building a factory because the anonymous function is now the mechanism by which we delay execution. It's not something I have done in apps, and maybe others don't use ExMachina like that, but I also don't know all the cases for how ExMachina is used in the wild.

What do you think Wil? Is it safe enough to use the anonymous function approach? I really do like it.

germsvel:
@WilHall I've given more thought to what you said, and it made me realize that perhaps we really can introduce this feature without adding a new function like `build_lazy/2`. We can just take anonymous functions. 

What I really like about the new approach is that we can also include passing the parent factory into the anonymous function so that we can get something like "derived" attributes. 

In any case, I'm opening a different PR since it's fairly different. 

Closing this in favor of https://github.com/thoughtbot/ex_machina/pull/408

PR #410, opened 2021-01-27
import evaluate_lazy_attributes for ExMachina macro

dvic:


dvic:
Is there a reason why `evaluate_lazy_attributes` is not automatically imported? If not, this PR should fix that :)

germsvel (approved):
Nice catch! I missed that. 🔍

Thanks for contributing. 🥳

Issue #273, opened 2018-02-09
Response of `insert` doesn’t contain default values


nallwhy:
After insert via Ecto, the response doesn't contain default values.

```elixir
# Migration
create table(:models) do
  add :date, :naive_datetime, null: false, default: fragment("now()")
end

# Model
schema "models" do
  field(:date, :naive_datetime)
end

Repo.insert(Model, %{})
=>
{:ok,
 %Model{
   __meta__: #Ecto.Schema.Metadata<:loaded, "models">,
   date: nil
}}
```

But when I load that model, it contains date.

```elixir
%Model{
   __meta__: #Ecto.Schema.Metadata<:loaded, "models">,
   date: ~N[2018-02-06 11:47:41.517655]
}
```
What should I do to get response from Ecto insert that contains default values?

tielur:
@nallwhy Ecto's `Repo.insert` doesn't return data by default because not all databases support it. If you are using a database that supports it, like Postgres, you could do something like this:

`model = build(:model) |> Repo.insert!(returning: true)`

That would build your model and insert it into the database, and then return the date field.

If you really wanted this functionality out of the box when call `insert(:model)` then you could write your own `ExMachina.Strategy`, pretty much take the built in [EctoStrategy](https://github.com/thoughtbot/ex_machina/blob/master/lib/ex_machina/ecto_strategy.ex) and modify it to suit your needs with `insert`.

Here's a working example:
I added a new [EctoStrategyWithData](https://gist.github.com/tielur/78b179de897d6659eef4a2db8a3e82c5)

Then with your Factories use that instead of the provided `ExMachina.Ecto`:
`use MyApp.EctoStrategyWithData, repo: MyApp.Repo`

Then your example above will work.

tielur:
Another different option is to set the default value in the Factory to something like: `NaiveDateTime.utc_now()`

```elixir
def model_factory do
  %ExMachinaInsertTest.Model{
    date: NaiveDateTime.utc_now()  
  }
end
```

sgtpepper43:
Your above solution doesn't _quite_ work, if I just do
```elixir
use MyApp.EctoStrategyWithData, repo: MyApp.Repo
```
Then I'm missing all the build functions, since that strategy only does the `:insert` method. I can do
```
use ExMachina
use MyApp.EctoStrategyWithData, repo: MyApp.Repo
```
[...]

Issue #382, opened 2020-05-19
There is no option to pass opts


mojidabckuu:
Never used ex_machina without Ecto under the hood. Today faced with a case when I want to pass *Shared options* to my Repo's insert function. Barely reviewing the code I found that the "core"  in `insert!`call doesn't use `opts` parameter
```elixir
record
|> cast
|> repo.insert!
```
and seems there is no way to pass it

So I am looking for something like this

```elixir
defmodule MyApp.Factory do
  use ExMachina.Ecto, repo: MyApp.Repo, opts: [foobar: :enable]
  ...
end
```
where later this opts will be passed as a part of the insert

germsvel:
Hi @mojidabckuu I think what you're looking for has an open PR, and we're looking for someone to add tests to it, if you're up for it! https://github.com/thoughtbot/ex_machina/pull/377

germsvel:
I just pushed v2.6.0 of ExMachina which includes passing Repo options to [`insert/3`](https://hexdocs.pm/ex_machina/2.6.0/ExMachina.Ecto.html#c:insert/3). Hope that helps with this issue!

PR #411, opened 2021-02-02
Pass opts to Repo.insert! (add function-level opts to strategies)

germsvel:
Closes: https://github.com/thoughtbot/ex_machina/issues/273

Related to: https://github.com/thoughtbot/ex_machina/issues/382, https://github.com/thoughtbot/ex_machina/issues/280, https://github.com/thoughtbot/ex_machina/pull/377

What changed?
=============

We add the ability for a strategy to receive more options as a third argument. These options could be a function-level override of the strategy's options, or some other function-level options that we want to pass to the underlying strategy implementation.

For example, the EctoStrategy can use those options and pass them to the `Repo.insert/2` function:

```elixir
insert(:user, [name: "gandalf"], returning: true)
build(:user, name: "gandalf") |> insert(returning: true)
```

That allows people to have more control over inserting the factory into the database. For example, we can now automatically get db-generated values by passing `returning: true`. Or we can use `prefix` or `on_conflict`.

For a full list of insert options, see [Repo.insert docs].

[Repo.insert docs]: https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2-options

Known less-than-good ergonomics
-------------------------------

Because both the second and third arguments are lists or maps, we need to add the surrounding keyword list `[]` to the second argument when we're using `insert/3`.

In the worst case, when we don't pass args to the factory but we pass options to the repo, we're required to define an empty list:

```elixir
insert(:user, [], returning: true)
```

We also have to wrap previously unwrapped lists _when_ we use the third opts argument:

```elixir
insert(:user, [name: "Aragorn"], returning: true)
```

That's just the nature of having two arguments that are both lists or maps (factory args and repo opts).

It reads more nicely when it's an already built struct:

```elixir
build(:user) |> insert(returning :true)
build(:user, name: "james") |> insert(returning :true)
insert(%User{}, returning: true)
```

Also available for [strategy_func]_pair and _list functions
-------------------------------------------------------------

We also update the strategy's _pair and _list function generators to allow for `opts` to be passed, since presumably users might want to do something like this:

```elixir
insert_pair(:user, [name: "jane"], returning: true)
```

Current strategies in the wild that don't support this will not break. If they implement a `handle_[function_name]/3`, then they can also accept the third arguments, and their users can pass the third argument. Otherwise, they can continue using the strategies as they already have.


lib/ex_machina/strategy.ex
```diff
@@ -66,16 +94,39 @@ defmodule ExMachina.Strategy do
...
+            unquote(function_name)(record, opts)
+          end
+
+          def unquote(function_name)(factory_name, attrs) do
             record = ExMachina.build(__MODULE__, factory_name, attrs)
 
             unquote(function_name)(record)
           end
 
+          def unquote(function_name)(factory_name) do
```
germsvel, line 109: We now handle each case directly in a separate function (instead of having optional `attrs \\ %{}`) because having the optional args conflicts with other versions of `insert/2` and `insert/3`.

priv/test_repo/migrations/1_migrate_all.exs
```diff
@@ -6,8 +6,28 @@ defmodule ExMachina.TestRepo.Migrations.MigrateAll do
...
+      RETURN NEW;
+    END;
+    $$;
+    """)
+
+    execute(~S"""
+    CREATE TRIGGER gen_db_value
+    BEFORE INSERT ON users
+    FOR EACH ROW
+    EXECUTE FUNCTION set_db_value();
```
germsvel, line 28: Created this function and trigger so that we could properly test `returning: true`. Maybe it's testing too much? But it seemed like a good test that we're indeed passing the values to `Repo.insert!/2` as it expects them.

README.md
```diff
@@ -404,6 +406,24 @@ end
 Using `insert/2` in factory definitions may lead to performance issues and bugs,
 as records will be saved unnecessarily.
 
+### Passing options to Repo.insert!/2
+
+`ExMachina.Ecto` uses
+[`Repo.insert!/2`](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert!/2) to
+insert records into the database. Sometimes you may want to pass options to deal
+with multi-tenancy or return some values generated by the database. In those
```
WilHall, line 414: How would you feel about being more explicit here about the use case for `returning`? Specifically it's not just values generated by the database, but values for all columns regardless of if those columns are specified in the factory (and by proxy, in the `insert`).

lib/ex_machina/ecto.ex
```diff
@@ -52,6 +52,25 @@ defmodule ExMachina.Ecto do
   @callback insert(factory_name :: atom) :: any
   @callback insert(factory_name :: atom, attrs :: keyword | map) :: any
 
+  @doc """
+  Builds a factory and inserts it into the database.
+
+  The first two arguments are the same as `c:ExMachina.build/2`. The last
+  argument is a set of options that can be passed to Ecto's
```
WilHall, line 59:
```suggestion
  argument is a set of options that will be passed to Ecto's
```

lib/ex_machina/ecto.ex
```diff
@@ -52,6 +52,25 @@ defmodule ExMachina.Ecto do
...
+  [`Repo.insert!/2`](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert!/2).
+
+  ## Examples
+
+      # return values from the database
```
WilHall, line 64:
```suggestion
      # return all values from the database
```

WilHall (approved):
🙌🏽

README.md
germsvel: I don't know that I want to go into full detail for what `returning` does. Rather, I was hoping to push people to look up what is available in the `Repo.insert!/2` docs. I just happened to know that multi-tenancy and returning are two things people have asked for. Do you have a suggestion on how to write this in a way that communicates that better? Maybe something like this?

> Sometimes you may want to pass options to Repo's insert function (e.g. `returning: true` or `prefix: "prefix"`). In those cases...

PR #412, opened 2021-02-03
Remove "web" dir from README.md

Apelsinka223:
Update due to current practice of directory structure


germsvel (approved):
Thank you for spotting that and making the change. 🎉

PR #422, opened 2022-05-25
Fix syntax error in ExMachina.Sequence docs

Stratus3D:


brimatteng:
duplicated here -> https://github.com/thoughtbot/ex_machina/pull/421

PR #426, opened 2022-11-12
Mix format to fix build error

cschilbe:
Build was failing after my previous PR was merged due to mix format being required.


btkostner:
Looks like formatting was fixed in https://github.com/beam-community/ex_machina/pull/439

PR #433, opened 2023-08-29
construct functions to be defined as well as {thing}_factory

byronanderson:
order of attempts:

{thing}_factory/2
{thing}_factory/1
construct/2
construct/1

(if construct/2 is defined, it will always be used instead of construct/1)

@jfeeny was interested in introspecting metadata attached to struct modules in order to provide sane defaults without needing manually to register `person_factory`, for example, you could do something like this:

```elixir
defmodule Factory do
  use ExMachina
  def construct(type) do
    struct = Module.camelize(type)
    fields = struct.describe() # __schema__ in ecto, for example
    fields
    |> Enum.map(fn {name, type} ->
      {name, default_for(type)}
    end)
    |> struct!(struct)
  end

  defp default_for(:string) do
    generate_fake_cat_name()
  end
end
```

### DIFF
--- a/lib/ex_machina.ex
+++ b/lib/ex_machina.ex
@@ -231,11 +231,19 @@ defmodule ExMachina do
     function_name = build_function_name(factory_name)
 
     cond do
-      factory_accepting_attributes_defined?(module, function_name) ->
+      function_defined?(module: module, name: function_name, arity: 1) ->
         apply(module, function_name, [attrs])
 
-      factory_without_attributes_defined?(module, function_name) ->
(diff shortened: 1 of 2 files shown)


doomspork:
Howdy @byronanderson! Now that ThoughtBot has passed this library on to us we're in the process of getting things updated 🎉 I was looking at your PR but I what I lack is the context for this change, can you point towards any issues or discussions around the need for these new functions? 

Separately, if we are going to add new capabilities I would like us to include documentation along with the changes.

PR #437, opened 2024-01-17
chore: sync files with beam-community/common-config

doomspork:


btkostner (approved):
DIalyzer shows a valid issue, but we can fix that in another PR

